Issue #144, opened 2018-02-07
New endpoint to provide data about all relays in the system, irrespective of workspace


sungo:
This should be pinned to Administrators in the GLOBAL workspace.

PR #147, opened 2018-02-07
/relay endpoint

sungo:
As per #144

### DIFF
--- a/Conch/lib/Conch/Model/Relay.pm
+++ b/Conch/lib/Conch/Model/Relay.pm
@@ -10,6 +10,9 @@ Conch::Model::Relay
 package Conch::Model::Relay;
 use Mojo::Base -base, -signatures;
 
+use Conch::Class::Relay;
+use Conch::Time;
+
 use Try::Tiny;
 
 has 'pg';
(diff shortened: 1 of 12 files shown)


Conch/lib/Conch/Model/Relay.pm
```diff
@@ -108,6 +111,25 @@ sub connect_device_relay ( $self, $device_id, $relay_id ) {
...
+=head2 list
+
+Provide a list of all relays in the database as Class::Relay objects
+
+=cut
+
+sub list ( $self ) {
+	my @relays;
+	try {
+		for my $r ($self->pg->db->query("select * from relay;")->hashes->@*) {
```
lseppala: You can use the SQL::Abstract interface, `...->db->select('relay')->...`

daleghent:
I looked over the schema and this seems to be exactly what I need!

sungo:
There are permissions tests in t/integration/04 to ensure that the readonly and integrator user classes can't access the relay list.

lseppala:
Ah, I see it now. The way GH had folded the code, it looked unrelated.

PR #149, opened 2018-02-13
Minor UI enhancements

bdha:
* Rename username field on login
* Remove RU column from rack list
* Link rack on device detail page


bdha:
@sungo @lseppala Can I get a review, please?

Conch/ui/src/views/Login.js
```diff
@@ -11,7 +11,7 @@ export default {
                 m("legend", t("Login to Conch")),
                 m("input[type=text]", {
                     oninput: m.withAttr("value", Auth.setUsername),
```
sungo: If the field is going to be the email address now, we probably should also change the variable names so that 'Auth.username' isn't secretly an email address.

lseppala: Captured in LP-43864914, I'll do the rename and make a new PR.

lseppala:
Renamed the variables and re-added the Joyent logo, which was accidentally removed.

Conch/ui/src/languages/ko.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
 
         /* Login items */
         "Login to Conch"          : "Conch 로그인",
-        "User Name"               : "사용자 이름",
+        "Login Email"             : "사용자 이름",
```
sungo: Probably should get an updated translation at some point

PR #156, opened 2018-02-15
Fix get and set single device setting

bdha:
We use device settings like `build.reboot_count`. Mojo translates this to:

```
key => build
format => reboot_count
```

This broke `conch-rebooter`, amongst other things using namespaced/URI-like settings.

I've tested this on the Mojo "prod" zone.

### DIFF
--- a/Conch/lib/Conch/Controller/DeviceSettings.pm
+++ b/Conch/lib/Conch/Controller/DeviceSettings.pm
@@ -37,7 +37,16 @@ overwritten
 
 sub set_single ($c) {
 	my $body          = $c->req->json;
-	my $setting_key   = $c->param('key');
+  my $setting_key;
+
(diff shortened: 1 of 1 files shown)


Conch/lib/Conch/Controller/DeviceSettings.pm
```diff
@@ -36,8 +35,18 @@ overwritten
 =cut
 
 sub set_single ($c) {
-	my $body          = $c->req->json;
-	my $setting_key   = $c->param('key');
+	my $body = $c->req->json;
```
sungo: I think there's a way to do this with less code by disabling extension based rendering. I'm investigating.

sungo:
We opted to make this change using an option in Mojo's content rendering system. See #157

PR #162, opened 2018-02-24
Validation System and new validation model

lseppala:
This add the basic Validation System components and converts legacy validations to use a new model. It includes abundant tests and documentation to assist others in building and verifying new validations.

See https://github.com/joyent/conch/blob/user/lane/validation-system/Conch/lib/Conch/Validation/README.md for introductory documentation on writing validations in the new model (generated from POD).

### DIFF
--- a/Conch/cpanfile.snapshot
+++ b/Conch/cpanfile.snapshot
@@ -2129,6 +2129,27 @@ DISTRIBUTIONS
       Fennec::Lite 0
       Test::Exception 0
       Test::More 0
+  Minion-8.12
+    pathname: S/SR/SRI/Minion-8.12.tar.gz
+    provides:
+      LinkCheck undef
+      LinkCheck::Controller::Links undef
+      LinkCheck::Task::CheckLinks undef
+      Minion 8.12
+      Minion::Backend undef
+      Minion::Backend::Pg undef
+      Minion::Command::minion undef
+      Minion::Command::minion::job undef
+      Minion::Command::minion::worker undef
+      Minion::Job undef
+      Minion::Worker undef
+      Minion::_Guard 8.12
+      Mojolicious::Plugin::Minion undef
+      Mojolicious::Plugin::Minion::Admin undef
+    requirements:
+      ExtUtils::MakeMaker 0
+      Mojolicious 7.56
+      perl 5.010001
   Mock-Quick-1.111
     pathname: E/EX/EXODIST/Mock-Quick-1.111.tar.gz
     provides:
@@ -3566,6 +3587,18 @@ DISTRIBUTIONS
       perl 5.006
       strict 0
       warnings 0
+  Submodules-1.0014
+    pathname: Z/ZA/ZARABOZO/Submodules-1.0014.tar.gz
+    provides:
+      Submodules 1.0014
+      Submodules::Result 1.0014
+    requirements:
+      ExtUtils::MakeMaker 0
+      File::Basename 0
+      File::Find 0
+      File::Spec 0
+      Test::More 0
+      perl 5.006
   Test-Deep-1.127
     pathname: R/RJ/RJBS/Test-Deep-1.127.tar.gz
     provides:
@@ -3686,10 +3719,10 @@ DISTRIBUTIONS
       perl 5.006
       strict 0
       warnings 0
-  Test-PostgreSQL-1.24
-    pathname: T/TJ/TJC/Test-PostgreSQL-1.24.tar.gz
+  Test-PostgreSQL-1.25
+    pathname: T/TJ/TJC/Test-PostgreSQL-1.25.tar.gz
     provides:
-      Test::PostgreSQL 1.24
+      Test::PostgreSQL 1.25
     requirements:
       DBD::Pg 0
       DBI 0
--- a/Conch/lib/Conch/Class/DatacenterRack.pm
+++ b/Conch/lib/Conch/Class/DatacenterRack.pm
@@ -22,9 +22,29 @@ with 'Conch::Class::Role::JsonV1';
 
 =head2 datacenter_room_id
 
+=head2 slots
+
+List of available slots defined by the rack layout
+
 =cut
 
-has [qw( id name role_name datacenter_room_id )];
+has [qw( id name role_name datacenter_room_id slots)];
+
+=head2 as_v1_json
+
+=cut
+
+sub as_v1_json {
+	my $self = shift;
+	return {
+		id => $self->id,
+		name => $self->name,
+		role_name => $self->name,
+		datacenter_room_id => $self->datacenter_room_id
+	};
+}
+
+1;
 
 1;
 
--- a/Conch/lib/Conch/Model/Validation.pm
+++ b/Conch/lib/Conch/Model/Validation.pm
@@ -85,7 +85,7 @@ sub upsert ( $class, $name, $version, $description, $module ) {
 		},
 		$name, $version, $description, $module
 	)->hash;
-	return $ret && $class->new( $ret->%* );
+	return $class->new( $ret->%* ) if $ret;
 }
 
 =head2 lookup
@@ -96,9 +96,9 @@ Lookup a validation by ID
 
 sub lookup ( $class, $id ) {
 	my $ret =
-		Conch::Pg->new->db->select( 'validation', $attrs,
+		Conch::Pg->new->db->select( 'validation', undef,
 		{ id => $id, deactivated => undef } )->hash;
-	return $ret && $class->new( $ret->%* );
+	return $class->new( $ret->%* ) if $ret;
 }
 
 =head2 list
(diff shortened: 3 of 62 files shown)


lseppala:
@sungo All good to review! Before you do, let's hop on a call and go over some of the design details and motivations.

Fortunately, most of the lines added are in tests.

Conch/lib/Conch/Validation.pm
```diff
@@ -0,0 +1,625 @@
...
+
+use Mojo::Base -base, -signatures;
+use Mojo::Exception;
+use JSON::Validator;
+use Try::Tiny;
+
+has 'name';
+has 'version';
+has 'description';
+has 'schema';
```
sungo: You'll see comments to this affect below but to put my larger thoughts in one place:

A lot of the validations in this PR don't use the schema to validate their input. They have custom logic inside the validation itself.  Is the JSON Schema not capable of validating those data sets? If it is capable, we need to transition those validations to use the Schema so that we don't, from day one, provide a common procedure of bypassing the Schema system entirely. If the Schema can't handle those data sets, I think we need to re-evaluate the Schema system. Basically, we need to pick a single data validation path, particularly for these first validations that will be held up as examples for future development.

Conch/lib/Test/Conch/Validation.pm
```diff
@@ -0,0 +1,285 @@
...
+			$case->{description} ? ' [' . $case->{description} . ']:' : ':';
+
+		diag("$msg_prefix input data: $pretty_data") if $debug;
+
+		$validation->clear_results;
+
+		if ( $case->{dies} ) {
+			diag("$msg_prefix should die") if $debug;
+			dies_ok( sub { $validation->run_unsafe($data) } )
+				|| diag "$msg_prefix Was expecting $validation_module "
```
sungo: why the diag here? `dies_ok` accepts a message parameter. https://metacpan.org/pod/Test::Exception#dies_ok

Conch/lib/Test/Conch/Validation.pm
```diff
@@ -0,0 +1,285 @@
...
+				description => 'Hello world test case',
+				debug       => 1
+			}
+		]
+	);
+
+
+=cut
+
+sub test_validation {
```
sungo: This is really hard to follow with all the diags and perltidy's odd manner of formatting long lines.

Conch/t/integration/04_test_datacenter_loaded.t
```diff
@@ -321,9 +322,55 @@ subtest 'Device location' => sub {
...
+	$t->delete_ok(
+		"/validation_plan/$validation_plan_id/validation/$validation_id")
+		->status_is(204);
+
+	$t->get_ok("/validation_plan/$validation_plan_id/validation")->status_is(200)
+		->content_is('[]');
+};
+
+
+$t->delete_ok('/device/TEST/location', 'can delete device location')->status_is(204);
```
sungo: Why did this test get moved down here?

Conch/t/model/Validation.t
```diff
@@ -0,0 +1,76 @@
...
+
+use_ok("Conch::Model::Validation");
+
+use Conch::Model::Validation;
+
+my $pgtmp = mk_tmp_db() or die;
+my $pg = Conch::Pg->new( $pgtmp->uri );
+
+my $validation;
+subtest "Create validation" => sub {
```
sungo: Are we just testing `has` here, that calling `create` sets attributes?

Conch/cpanfile.snapshot
```diff
@@ -118,16 +118,18 @@ DISTRIBUTIONS
       ExtUtils::MakeMaker 0
```
sungo: This cpanfile.snapshot lowers the version on a  lot of modules. Mind getting an update?

Conch/lib/Conch/ValidationSystem.pm
```diff
@@ -0,0 +1,140 @@
...
+use Mojo::Base -base, -signatures;
+use Minion;
+use Mojo::Log;
+use Mojo::Exception;
+use Submodules;
+
+use Conch::Model::ValidationState;
+use Conch::Pg;
+
+has 'minion' => sub { Minion->new( Pg => Conch::Pg->new ) };
```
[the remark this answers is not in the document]
lseppala: 👍

Conch/lib/Conch/Class/DatacenterRack.pm
```diff
@@ -22,9 +22,29 @@ with 'Conch::Class::Role::JsonV1';
...
+
+=cut
+
+has [qw( id name role_name datacenter_room_id slots)];
+
+=head2 as_v1_json
+
 =cut
 
-has [qw( id name role_name datacenter_room_id )];
```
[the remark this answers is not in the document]
sungo: Sigh.

Conch/lib/Conch/Model/ValidationState.pm
```diff
@@ -0,0 +1,76 @@
...
+Mark the Validation State as completed
+
+=cut
+
+sub mark_completed ( $self ) {
+	my $completed = Conch::Pg->new->db->update(
+		'validation_state',
+		{ completed => 'NOW()' },
+		{ id        => $self->id },
+		{ returning => 'completed' }
```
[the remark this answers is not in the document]
sungo: groovy

Conch/lib/Conch/Validation.pm
```diff
@@ -0,0 +1,625 @@
...
+	use Mojo::Base 'Conch::Validation';
+
+	sub validate {
+		my ($self, $data) = @_;
+		$self->register_result({ expected => 1, got => $data->{pass} });
+	}
+
+=cut
+
+sub validate ( $self, $data ) {
```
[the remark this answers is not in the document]
lseppala: I originally intended to make it abundantly clear to someone writing a validation for the first time. It blows up and they're told what they need to fix it.

I also use it as a hook to satisfy pod.t. With the `Pod::Coverage::CountParents` option that I set, Test::Pod::Coverage is satisfied if a method overridden in the child is present in the parent. If `validate` isn't defined here, it will be required. I don't want to require pod docs beyond copyright in the validations. The description fulfills describing what the validation is and does.

Conch/lib/Conch/Validation.pm
```diff
@@ -0,0 +1,625 @@
...
+
+Stop execution of the Validation immediately and record an error.
+
+	$self->die('This validation cannot continue!') if $bad_condition;
+
+=cut
+
+sub die ( $self, $message, $level = 1 ) {
+
+	die Mojo::Exception->new($message)->trace($level);
```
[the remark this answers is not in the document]
lseppala: My intention is that validations should die fast and loudly. Exceptions are caught and made into 'error' validation results and are reported in the API. I want to minimize ambiguity as much as possible. I want to avoid unexpected undefs cast into empty string in comparisons and validations that run but don't produce validation results because of a conditional branch. Dying fast also helps to be explicit about the inputs when writing the validation unit tests.

I see your point about the methods that immediately die if they have undefined data, like `device_location`, `hardware_product_profile`, etc. I'm not sure will be a problem under normal operation. We can expect hardware product always to be present when validation plans are executed by the orchestration system because it's a necessary component of a workflow. If a workflow step is executed before a device may have a device location, then it should be associated with a validation plan of validations that don't require a location.

I'll go ahead and add a `has_device_location` method to return a boolean in the case that there's some conditional based on whether a device has an assigned location. I don't there will be many validations that use this type of logic (and they might be better split into two validations in separate pre- and post-location plans), but it's possibly useful.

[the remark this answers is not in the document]
sungo: See also my comment in https://github.com/joyent/conch/pull/162#discussion_r176791133

Conch/lib/Test/Conch/Validation.pm
sungo: I'd prefer that we use the native syntax, since the only reason for variation is a formatting nit. Test names show up in the TAP reports when failures occur and are functionally different than diagnostic output.

Conch/lib/Conch/Model/Validation.pm
```diff
@@ -96,19 +97,32 @@ Lookup a validation by ID
 
 sub lookup ( $class, $id ) {
 	my $ret =
-		Conch::Pg->new->db->select( 'validation', undef,
+		Conch::Pg->new->db->select( 'validation', $attrs,
```
lseppala: Reset back to use $attrs.

Using `undef` returns fields not purposely exposed in the module. These fields are stored in the internal blessed hashref. These fields cause `is_deeply` tests to fail when comparing whether an object created with `->create` matches an object created with `->lookup`.

I like the semantics of specifying the attributes on each query. It's decouples the model from the representation in the database. The schema may change, even be split across multiple tables, and the attributes can remain the same.

sungo: Ah. So `Mojo::Base` is annoyingly minimal here. Over in `Moo` land, if you call `new()` with random attributes, they get ignored.

PR #163, opened 2018-02-26
fix: change GET /version to be statically configurable

lseppala:
The value will be set by configuration management (Ansible) when
deploying new versions. This removes any need for git to be available or
in the path

A corresponding change has been made in our infrastructure repo (link withheld).

### DIFF
--- a/Conch/conch.conf.dist
+++ b/Conch/conch.conf.dist
@@ -6,6 +6,9 @@
     # URI format is postgresql://[user[:password]@][netloc][:port][/dbname][?param1=value1&...]
     pg => 'postgresql://conch@/conch',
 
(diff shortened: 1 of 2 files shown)


sungo:
Separating the version out from the API entirely makes testing version-sensitive features difficult in dev, given that the a dev revision will now always have a version of 'v2.0.0-dev'. I added the version endpoint to allow clients to determine which functions are supported, including against development versions that were not deployed by Ansible. As of this PR, dev versions require custom hand-maintained config files so that their versions are usable for client testing.

bdha:
Given that, I'm happy to fix the SMF method's PATH and revert this.

sungo:
With #164 merged in, I'm going to revert this PR.

PR #164, opened 2018-02-27
Move conch exec method to script

bdha:
This moves the SMF method from a direct call in the manifest to a script, making it simpler to change the service's behavior. (We no longer need to reimport the manifest when we update something.) 

Anecdotally, I also find it easier to reason about a shell script that commands wedged into the middle of an XML document.

I've tested import, stop, and restart. All seem fine.

Conch no longer complains about being unable to find `git`, as well.

### DIFF
--- a/smf/conch.xml
+++ b/smf/conch.xml
@@ -19,24 +19,11 @@
         </dependency>
 
 
-        <method_context>
-
-        </method_context>
-<exec_method type="method" name="start" exec="/opt/local/lib/perl5/site_perl/bin/carton exec hypnotoad bin/conch" timeout_seconds="60">
-            <method_context working_directory='/opt/conch/Conch'></method_context>
-        </exec_method>
-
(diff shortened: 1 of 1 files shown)


bdha:
This was something I would have done anyway, but #163 was the impetus.

lseppala:
Why isn't this script in version control?

Also, refresh != kill. Refresh uses hypnotoad's zero-downtime deployment, and stop kills the process. Also, hypnotoad daemonizes and exits immediately.

Looking at the script on conch-api1, it is broken. It will not stop or refresh. Deploying new goes is not possible with Ansible.

bdha:
@lseppala gah, sorry. My fault. I'll fix this another PR shortly!

But: SMF `:kill` does do the correct thing for stop and refresh -- it just doesn't use the hypnotoad zerodowntime bit, which I'll fix. SMF uses `contract(4)` to track processes to do the right thing.

PR #170, opened 2018-02-28
Add SMF method script for conch

bdha:
Move SMF exec method to a script. See #164 for the first pass at this.

Supports hypnotoad hotload when running `svcadm refresh conch`.

### DIFF
--- a/smf/conch
+++ b/smf/conch
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+export PATH=$PATH:/opt/local/bin:/opt/local/sbin
+cd /opt/conch/Conch
+
+case "$1" in
+  start)
+    /opt/local/lib/perl5/site_perl/bin/carton exec hypnotoad bin/conch
+    ;;
+  stop)
+    /opt/local/lib/perl5/site_perl/bin/carton exec hypnotoad -s bin/conch
+    ;;
+  restart)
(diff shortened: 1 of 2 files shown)


smf/conch
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+export PATH=$PATH:/opt/local/bin:/opt/local/sbin
+cd /opt/conch/Conch
+
+case "$1" in
+  start)
+    /opt/local/lib/perl5/site_perl/bin/carton exec hypnotoad bin/conch
```
lseppala: Put this into the path too?

smf/conch.xml
```diff
@@ -19,9 +19,9 @@
         </dependency>
 
 
-        <exec_method name='start' type='method' exec='/opt/conch/smf/conch' timeout_seconds='60'/>
+        <exec_method name='start' type='method' exec='/opt/conch/smf/conch start' timeout_seconds='60'/>
         <exec_method name='stop' type='method' exec=':kill' timeout_seconds='60'/>
```
lseppala: Use the `stop` via the script, which kills workers gracefully

bdha: Done.

smf/conch
bdha: Done.

PR #180, opened 2018-03-01
feat: add Conch::Time->now()

sungo:


Conch/lib/Conch/Time.pm
```diff
@@ -80,7 +104,10 @@ B<NOTE:> This method will negatively impact performance if called frequently.
 =cut
 
 sub to_datetime {
-	return $_parser->parse_datetime( shift->timestamp );
+	return DateTime::Format::Strptime->new(
```
lseppala: I pulled this out into a static variable because I figured we only ever need one instance. Mind helping me understand why it's better to do it this way?

sungo: It's cleaner. Global variables, in general, give me hives. It could be moved into a proper constant that would allow the compiler to inline it but that feels a lot like premature optimization.

PR #189, opened 2018-03-06
docs: reinstate the openapi docs. 

sungo:
they will need to be updated for changes since January


lseppala:
You had made the case to me that the docs were confusing, not very useful, and harmful in their inaccuracy, which is why I removed them. We agreed that public consumption and documentation would be better served with the CLI. Why are these being brought back? What is the expectation for maintenance?

sungo:
The short version is that our customers changed my mind and impressed upon me the importance of the docs we had before. I figured that bringing these back was a better route than trying to implement or write something new.

lseppala:
Going to make some changes (using yarn instead of npm, updating some deps) in a separate PR

PR #197, opened 2018-03-08
build: travis integration

sungo:
# Basics
- See https://travis-ci.org/joyent/conch for results
- Travis only supports Linux for perl apps. So it won't catch any SmartOS build issues.
- It will build against PRs so, as much as I like the WIP PRs, we might need to do something else. I'm not sure everyone wants build checks everytime we push to WIP branches.
- Builds can also be triggered manually in the UI or their CLI

# Dependencies
- Carton dependencies are cached against the branch they're on. So the first time one builds against a branch, it will take a while because carton needs to install deps. 
- Dependency caching means that if one adds a new module, caches will need to manually invalidated at https://travis-ci.org/joyent/conch/caches

# Releases
- I've not tested the release process in this PR but it's the same as joyent/conch-shell so it should work. I'll test it against master once the PR is merged.
- Github releases will be built for any new tag. 
- The release process is using an API token of mine. So, all releases generated by Travis will look like I did them.
- New releases will be marked as regular releases. If you want a pre-release, you'll need to flip that switch manually in the Github UI.

# Misc
- It is possible to also run tests against the UI but I've not set that up in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ install:
   - "cd Conch; carton install"
 script:
   - "cd Conch; carton exec prove -lpr t/"
+cache:
+  directories:
+    - "Conch/local"
 deploy:
   skip_cleanup: true
   provider: releases
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,12 @@ deploy:
   on:
     tags: true
     repo: joyent/conch
+notifications:
+  recipients:
+    - sungo@joyent.com
+      #- lane.seppala@joyent.com
+      #- bdha@joyent.com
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,33 @@
+os: linux
+sudo: false
+language: perl
+perl:
+    - "5.26"
+before_install:
+    - cpanm Carton
+install:
+  - "cd Conch; carton install"
```
lseppala: Can this use --deployment, so we're installing the versions in the lockfile?

sungo: Not until String::CamelCase fixes its META file

Issue #199, opened 2018-03-08
Failed validation criteria does not present a server as "FAIL" in the device and rack views


daleghent:
We are seeing cases where a server will fail validation due to something, but not be marked as such in the rack view for the device, or on the device's own page as it used to be.

Currently, https://conch.joyent.us/#!/device/S309573X8224784 is one such example. Validation is failing on the system's SMBIOS name, yet it is presented as as passing system.

sungo:
This appears to have been resolved by #202. Closing out the issue.

PR #202, opened 2018-03-09
 fix: catch unhandled exceptions in legacy validate and mark as fail

lseppala:
While investigating https://github.com/joyent/conch/issues/199, I discovered that any exceptions raised in the legacy validation code would be silently swallowed and not reported. This makes the following changes:

* It logs the exception and location in the main Mojo log
* It adds a failing device_validate record with the same exception message. This will show up in the UI.
* It marks the device's health as `FAIL`

Sidenote: I cannot wait to purge all this code. _soon.jpg_


Conch/lib/Conch/Legacy/Control/Device/Environment.pm
```diff
@@ -55,26 +47,26 @@ sub validate_cpu_temp {
 
 		my $method = "${cpu}_temp";
 
-		if ( $device_env->$method > $criteria->crit ) {
+		if ( $device_env->$method > 70 ) {
```
lseppala: All of these values have remained unchanged in the database. They will likewise be code-specified values in the new validation system.

PR #206, opened 2018-03-22
Report exceptions to Rollbar

lseppala:
Adds support to report exceptions to Rollbar. Reported details include stack traces, the affected user (if authenticated), code version (via current SHA), and server info. All of this information is used by Rollbar to help provide introspection.

I'll send out Rollbar invites to everyone.

### DIFF
--- a/Conch/cpanfile.snapshot
+++ b/Conch/cpanfile.snapshot
@@ -283,6 +283,7 @@ DISTRIBUTIONS
       Config::Any::XML undef
       Config::Any::YAML undef
     requirements:
+      Config::General 2.47
       Module::Pluggable::Object 3.6
   Config-General-2.63
     pathname: T/TL/TLINDEN/Config-General-2.63.tar.gz
@@ -368,8 +369,8 @@ DISTRIBUTIONS
       Test::More 0.47
       Tie::Hash 0
       perl 5.006
-  DBI-1.640
-    pathname: T/TI/TIMB/DBI-1.640.tar.gz
+  DBI-1.641
+    pathname: T/TI/TIMB/DBI-1.641.tar.gz
     provides:
       Bundle::DBI 12.008696
       DBD::DBM 0.08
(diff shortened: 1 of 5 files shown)


Conch/lib/Conch/Plugin/Rollbar.pm
```diff
@@ -0,0 +1,112 @@
...
+		context => {
+			pre  => $exception->lines_before,
+			post => $exception->lines_after
+		}
+	};
+
+	# We only have context for the first frame
+	$frames[0]->{context} = $context;
+
+	my $version = `git rev-parse HEAD`;
```
sungo: I'd like to avoid executing git for every exception we encounter. I'd rather put this into a variable on startup.

lseppala: I don't think performance is too much of a concern.
```
❯ perl -e'
use Benchmark q/timethis/;

timethis(1_000, q/`git rev-parse HEAD`/);
'
timethis 1000:  3 wallclock secs ( 0.06 usr  0.27 sys +  1.27 cusr  0.58 csys =  2.18 CPU) @ 458.72/s (n=1000)
```

We shouldn't be throwing so many exceptions that this code path is executed frequently in any case. I don't think how responsively a user gets their impending 500 'Something went wrong' is a big issue.

If still concerned, let's gather all instances of git executions and put them in the same plugin ( `Conch::Plugin::Git` or w/e) in a separate PR so the indirections are all in the same place.

lseppala: @sungo Went ahead and made the change by pulling it up to the register method.

I threw together a GitVersion plugin to pull all the git-version-y things in one place. If you thing it's worthwhile, I'll throw it in a PR.
https://github.com/joyent/conch/compare/user/lane/git-version-plugin

lseppala: Ok, wasn't sure. Will put the plugin in a separate PR (and update the other whichever gets merged first).

lseppala:
Updated to use GitVersion plugin

Conch/cpanfile.snapshot
```diff
@@ -118,16 +118,18 @@ DISTRIBUTIONS
       ExtUtils::MakeMaker 0
       Scalar::Util 0
       perl 5.006
-  Class-C3-Componentised-1.001002
```
sungo: This is rolling back module versions. Mind updating?

lseppala: Blew away `local` and ran carton install. No longer downgrading things. Pushed change.

PR #211, opened 2018-03-23
in status helper, add content negotiation and some convenience handling for 501 and 303 

sungo:


lseppala:
Do you want to replace the current usage `$c->status(303)` with this form in this PR or in another?

sungo:
I'd replace the current usage separately.

PR #216, opened 2018-03-28
Put auditing and rollbar into the feature flags system

sungo:


Conch/conch.conf.dist
```diff
@@ -18,7 +14,14 @@
     # rollbar_environment => 'staging',
 
 	features => {
+		'audit' => 0,
 		#'orc' => 1,
+		'rollbar' => 1,
```
lseppala: Let's make the rollbar 0 in .dist, and uncomment orc and set as default off (or on, whichever you want).

lseppala: Oh, also mind running perltidy on this file. Mix of spaces and tabs

sungo: Will do.

sungo: Done.

Issue #150, opened 2018-02-13
Orchestration System


sungo:
Conch needs a centralized system for controlling various processes inside the Conch environment, ranging from centralized management of production datacenter inventory validation to localized management of Conch relays and livesys images inside an integration facility.

In the most general terms, the orchestration system should be a purpose-built workflow engine. It must provide a way to specify policies, focused on build operations, that determine the order of operations for the Conch relay and livesys applications. The validation system must be used to determine if those policies have completed successfully.

The user interface, whether HTML or CLI, is *not* described in this RFD and will be specified at a later point.

# Concepts

## Automation

Automation is the execution of individual tasks, with the goal of simplifying or standardizing tasks that were often run manually previously.

## Workflow

A workflow is a set of pre-defined automations launched by a trigger condition. In the most general case, steps in a workflow may be optional or required and it is usually possible to nest workflows.

## Workflow Engine

A workflow engine is a system that evaluates trigger conditions, trigger individual workflows, and processes the results. Usually workflow engines are purpose-built for the needs of a specific application.

## Orchestration

Orchestration is concerned with bringing workflows together into processes or policies, with the goal of streamlining and reusing those processes. Automation provides the building blocks upon which orchestration processes are built.

# Design

## Workflows

Orchestration operations are tied to a 'workflow', which itself is tied to a hardware product. Every device is keyed to a hardware product, as well. This combination allows for workflow steps and validations to be linked to specific hardware revisions.

This is particularly useful when a newer version of a server design varies wildly from an older one. Both systems can be built and validated because their workflow, and the validations necessary to green-light the device, are tied to the hardware product specification.

## Workflow Steps

Workflow steps are an ordered list of string names and validations. The string names are opaque to the orchestration system but signal different operations to the downstream clients. 

## Status

Two types of status exist, one for the execution of an entire workflow, and another for the execution of an individual step.

### Workflow Status

Workflow status records the state of the execution of an entire workflow for a specific device. Most of the time, a device's workflow will either be "ongoing" or "completed". However, special circumstances may arise with a need to interrupt that flow.

If an external entity (probably a human) determines that a workflow must be stopped, the workflow status for that device will be set to 'abort'. The engine cannot reach out to client devices and forcibly halt execution so 'abort' indicates that workflow must cease when the current step finishes. When the
device completes its current step, the workflow status for the device will be set to 'stopped'.

[...]

PR #219, opened 2018-04-02
Device Roles and Services: defining the intended usage of a device

sungo:


Conch/lib/Conch/Controller/DeviceRoles.pm
```diff
@@ -0,0 +1,177 @@
...
+Get a single device service
+
+=cut
+
+sub get_one ($c) {
+	my $s = Conch::Model::DeviceRole->from_id($c->param('id'));
+	return $c->status(404 => { error => "Not found" }) unless $s;
+	return $c->status(404 => { error => "Not found" }) if $s->deactivated;
+
+	$c->status(200, $s->TO_JSON);
```
lseppala: Out of curiosity, is the call to `TO_JSON` necessary? Mojo::JSON should invoke it if present.

Conch/lib/Conch/Model/DeviceRole.pm
```diff
@@ -0,0 +1,361 @@
...
+
+
+
+=head2 from_name
+
+	my $role = Conch::Model::DeviceRole->from_name($name);
+
+Returns a DeviceRole given an existing name
+
+Includes deactivated objects
```
lseppala: Including deactivated means the query can return 0-or-many results, but this method only returns the first of those. You could end up in a situation where you've deactivated a role, created a new one with the same name, and this method would only return the deactivated role rather than the new one.

Conch/lib/Conch/Model/DeviceService.pm
```diff
@@ -0,0 +1,230 @@
...
+
+
+
+=head2 from_name
+
+	my $service = Conch::Model::DeviceService->from_name($name);
+
+Returns a DeviceService given an existing name
+
+Includes deactivated objects
```
lseppala: Same issue as mentioned above. Deactivated and recreating with the same name may only return the deactivated one.

Conch/t/integration/04_test_datacenter_loaded.t
```diff
@@ -21,7 +20,8 @@ my $dbh = DBI->connect( $pgtmp->dsn );
 my $t = Test::Mojo->new(
 	Conch => {
 		pg      => $pgtmp->uri,
-		secrets => ["********"]
+		secrets => ["********"],
+		features => { rollbar => 0 },
```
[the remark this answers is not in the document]
sungo: I disabled this because the exception handler in the rollbar plugin was getting called and hanging the tests.  This is what led to PR #218 so I could make sure the rollbar code wasn't going to get called at all in the tests.

sql/migrations/0024-device-roles.sql
```diff
@@ -0,0 +1,36 @@
...
+	updated     timestamptz not null default current_timestamp,
+	deactivated timestamptz
+);
+create unique index on device_service(name) where deactivated is null;
+
+
+create table device_role (
+	id                  uuid primary key default gen_random_uuid(),
+	name                text not null,
+	hardware_product_id uuid not null references hardware_product(id),
```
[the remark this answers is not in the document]
sungo: Let's have a call. The whole point of device roles was to create this structure of tying hardware and software together.

Conch/lib/Conch/Controller/DeviceRoles.pm
```diff
@@ -0,0 +1,177 @@
...
+	return $c->status(403) unless $c->is_global_admin;
+
+	my $body = $c->req->json;
+	if($body->{id}) {
+		return $c->status(400 => { error => "'id' parameter not allowed'"});
+	}
+	return $c->status(400 => { error => "'name' parameter required"})
+		unless $body->{name};
+
+	return $c->status(400 => {
```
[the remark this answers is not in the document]
sungo: I'm going to roll a more generic solution into a follow-on PR that'll let us use this sort of validation in all the controllers, like we do for the tests.

sql/migrations/0024-device-roles.sql
```diff
@@ -0,0 +1,35 @@
...
+	id                  uuid primary key default gen_random_uuid(),
+	name                text not null,
+	description         text,
+	hardware_product_id uuid not null references hardware_product(id),
+
+	created     timestamptz not null default current_timestamp,
+	updated     timestamptz not null default current_timestamp,
+	deactivated timestamptz
+);
+create unique index on device_role(name) where deactivated is null;
```
bdha: How will this handle instances where we have headnodes or admin servers that might be on R510s or R730s or whatever, depending on the datacenter?

Should this be an index between name and hardware_product, so we can account for

compute_node hw1
compute_node hw2
headnode hw1
headnode hw3
manta_storage shrimp_v3
manta_storage shrimp_v4

Or am I misunderstanding?

sungo: The 'name' field is intended for use in the UIs and tools to present a list of roles. It has no internal semantic meaning. "headnode on a r510" is a role where the hardware product is "r510" and the service list contains an entry for "headnode".

I'm going to drop the name field entirely, I think, to prevent folks from thinking that creating a role with a text name of "triton - r730" makes it so. In the tools, we'll use the description field and build "names" using a combination of the hardware profile and device service name fields.

sungo: Done.

sungo:
Part of #150

PR #220, opened 2018-04-03
feat: add Conch::Plugin::JsonValidator, allowing controllers to optionally validate input and output.

sungo:


Conch/lib/Conch/Plugin/JsonValidator.pm
```diff
@@ -0,0 +1,139 @@
...
+	});
+
+	####
+
+	my $output_validator = JSON::Validator->new();
+	$output_validator->schema(OUTPUT_SCHEMA_FILE);
+	_add_uuid_validation($output_validator);
+
+	$app->helper(
+		status_with_validation => sub ($c, $status_code, $data, $schema) {
```
lseppala: How about flipping the data and the schema, so you can drop the quotes? Looks pretty nice 
```
$c->status_with_validation(200, Workflow => $workflow):
```

sungo: Done.

PR #221, opened 2018-04-04
feat: Orc Lifecycles, an ordered list of workflows, bound to a device role

sungo:
Part of #150

### DIFF
--- a/Conch/json-schema/v1.yaml
+++ b/Conch/json-schema/v1.yaml
@@ -871,7 +871,7 @@ definitions:
       - deactivated
       - locked
       - steps
-      - product_id
+      - preflight
     properties:
       deactivated:
         anyOf:
@@ -879,9 +879,6 @@ definitions:
           - type: string
             format: date-time
       id:
-        type: string
-        format: uuid
-      product_id:
         type: string
         format: uuid
       name:
@@ -895,6 +892,8 @@ definitions:
         type: string
         format: date-time
       locked:
+        type: number
+      preflight:
         type: number
       steps:
         anyOf:
@@ -1032,3 +1031,43 @@ definitions:
     type: array
     items:
       $ref: "#/definitions/DeviceService"
+  OrcLifecycle:
+    type: object
+    additionalProperties: false
+    required:
+      - id
+      - name
+      - version
+      - locked
+      - role_id
+      - plan
+      - created
+      - updated
+    properties:
+      id:
+        type: string
+        format: uuid
+      name:
+        type: string
+      created:
+        type: string
+        format: date-time
+      updated:
+        type: string
+        format: date-time
+      version:
+        type: number
+      role_id:
+        type: string
+        format: uuid
+      locked:
+        type: number
+      plan:
+        type: array
+        items:
+          type: string
+          format: uuid
+  OrcLifecycles:
+    type: array
+    items:
+      $ref: "#/definitions/OrcLifecycle"
--- a/Conch/lib/Conch/Controller/Orc/Lifecycles.pm
+++ b/Conch/lib/Conch/Controller/Orc/Lifecycles.pm
@@ -1,0 +1,134 @@
+=head1 NAME
+
+Conch::Controller::Orc::Lifecycles
+
+=head1 METHODS
+
+=cut
+
+package Conch::Controller::Orc::Lifecycles;
+
+use Mojo::Base 'Mojolicious::Controller', -signatures;
+
+use Conch::Orc;
+
+
+=head2 get_all
+
+Get all lifecycles
+
+=cut
+
+sub get_all ($c) {
+	return $c->status_with_validation(200, 
+		OrcLifecycles => Conch::Orc::Lifecycle->all()
+	);
+}
+
+
+=head2 get_one
+
+Get a single lifecycle
+
+=cut
+
+sub get_one ($c) {
+	my $l = Conch::Orc::Lifecycle->from_id($c->param('id'));
+	return $c->status(404 => { error => "Not found" }) unless $l;
+
+	$c->status_with_validation(200, OrcLifecycle => $l );
+}
+
+
+=head2 create
+
+Create a new lifecycle
+
+=cut
+
+sub create ($c) {
+	return $c->status(403) unless $c->is_global_admin;
+	my $body = $c->validate_input('OrcLifecycleCreate') or return;
+
+	$body->{version} = 0 unless $body->{version};
+	if(Conch::Orc::Lifecycle->from_name_and_version(
+		$body->{name},
+		$body->{version}
+	)) {
+		return $c->status_with_validation(400, Error => {
+			error => "Lifecycle already exists with this name and version"
+		});
+	}
+
+	
+	my $l = Conch::Orc::Lifecycle->new($body->%*)->save();
+
+	$c->status(303 => "/o/lifecycle/".$l->id);
+}
+
+
+=head2 add_workflow
+
+Add a workflow to a lifecycle plan, with an optional plan_order 
+
+=cut
+
+sub add_workflow ($c) {
+	return $c->status(403) unless $c->is_global_admin;
+
+	my $l = Conch::Orc::Lifecycle->from_id($c->param('id'));
+	return $c->status(404 => { error => "Not found" }) unless $l;
+
+	my $body = $c->validate_input('OrcLifecycleAddWorkflow') or return;
+	
+	my $w = Conch::Orc::Workflow->from_id($body->{workflow_id});
+	return $c->status(404 => { error => "Workflow not found" }) unless $w;
+
+
+	if ($body->{plan_order}) {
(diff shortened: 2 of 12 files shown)


sql/migrations/0025-orc-lifecycles.sql
```diff
@@ -0,0 +1,29 @@
...
+	name        text        not null,
+	version     int         not null default 1,
+	created     timestamptz not null default current_timestamp,
+	updated     timestamptz not null default current_timestamp,
+	deactivated timestamptz,
+	locked      bool        not null default false,
+	role_id     uuid        not null references device_role(id),
+	unique(name, version)
+);
+create unique index on workflow_lifecycle(name) where deactivated is null;
```
lseppala: With the two unique indices on this table, do I follow that:

* Only one lifecycle with a given name can be active at a given time
* There can be multiple versions of the lifecycle with a given name, but older ones must be deactivated

Conch/lib/Conch/Controller/Orc/Lifecycles.pm
```diff
@@ -0,0 +1,128 @@
...
+
+Create a new lifecycle
+
+=cut
+
+sub create ($c) {
+	return $c->status(403) unless $c->is_global_admin;
+	my $body = $c->validate_input('OrcLifecycleCreate') or return;
+	
+	my $l = Conch::Orc::Lifecycle->new($body->%*)->save();
```
lseppala: You should check that the supplied name & version don't exist before attempting to create to avoid throwing 500s.

Conch/lib/Conch/Orc/Lifecycle.pm
```diff
@@ -0,0 +1,472 @@
...
+	try {
+		$ret = Conch::Pg->new->db->query(qq|
+			select l.*, array(
+				select wlp.workflow_id
+				from workflow_lifecycle_plan wlp
+				where wlp.lifecycle_id = l.id
+				order by wlp.plan_order
+			) as plan
+			from workflow_lifecycle l
+			where l.deactivated is null and l.id = ?;
```
lseppala: You provide $name but are querying by id

Conch/lib/Conch/Orc/Lifecycle.pm
```diff
@@ -0,0 +1,472 @@
...
+				order by l.version
+				limit 1
```
lseppala: Only one version of a name can be active based on the current schema. This bit is unnecessary unless the schema is changed.

Conch/lib/Conch/Orc/Lifecycle.pm
```diff
@@ -0,0 +1,472 @@
...
+}
+
+
+=head2 many_from_name
+
+Load all Lifecycles that match a given name
+
+=cut
+
+sub many_from_name ($class, $name) {
```
lseppala: Based on the current schema, only one version of a given name can be active at a given time.

Conch/json-schema/v1.yaml
```diff
@@ -1032,3 +1031,43 @@ definitions:
...
+        type: number
+      plan:
+        type: array
+        items:
+          type: string
+          format: uuid
+  OrcLifecycles:
+    type: array
+    items:
+      $ref: "#/definitions/OrcLifecycles"
```
lseppala: Should be singular.

Conch/lib/Conch/Orc/Lifecycle.pm
```diff
@@ -0,0 +1,472 @@
...
+
+=cut
+
+sub save ($self) {
+	my $db = Conch::Pg->new()->db;
+	$self->_set_updated(Conch::Time->now);
+
+	my %fields = $self->%*;
+	delete $fields{id};
+	delete $fields{plan};
```
lseppala: Should deactivated be allowed to be set? Created and updated should be managed by Postgres.

Conch/lib/Conch/Orc/Lifecycle.pm
```diff
@@ -0,0 +1,472 @@
...
+
+
+# Blows away the existing join records and recreates them using the array 
+# indices on ->plan as the plan_order field
+sub _rebuild_plan ($self) {
+	return $self unless $self->id;
+	my $db = Conch::Pg->new->db;
+	my $tx = $db->begin;
+
+	my @p = uniq($self->plan->@*);
```
lseppala: If there's duplicates in the plan, this will drop the latter duplicates. This may have surprising results. If I add workflow ID 1 as order 2 but workflow ID 1 is already a member of the plan as order 4, the order 4 plan will be silently dropped.

This is only an issue for adding new workflows. You may want to check if a workflow already exists in the add_workflow controller and return 409 if so. To reorder workflows, the user or client should explicitly remove and re-add the workflow at the desired location.

Conch/lib/Conch/Orc/Lifecycle.pm
```diff
@@ -0,0 +1,472 @@
...
+		}
+	}
+
+	my $ret;
+	try {
+		if($self->id) {
+			$ret = $db->update(
+				'workflow_lifecycle',
+				\%fields,
+				{ id => $self->id },
```
lseppala: set `updated` to `now()`

sungo: happens at the top of the routine via `$self->_set_updated(Conch::Time->now);`

Conch/json-schema/v1.yaml
sungo: I grow increasingly confused as to how schema based tests are passing with typos like this. For instance, the code uses this schema definition in two places in `#all` and the test for `#all`, both of which work just fine.

Conch/lib/Conch/Orc/Lifecycle.pm
sungo: The DB layer should be fixed for this use case now. The plan modification routines now check if the workflow is or isn't in the list and don't make a trip to the database if things are not as expected. It shouldn't be possible, with those changes, to get duplicate workflows into the plan.

Conch/lib/Conch/Orc/Lifecycle.pm
lseppala: I think Postgres should set this value. Also, on insert the created timestamp will be _later_ than the updated timestamp as it is.

sungo: This is an idiom that exists through orchestration and the device role code. I'm open to changing it but I just wanted to point out that it's not exactly new.

lseppala: Ah, I must have missed it. The updated timestamp being earlier than created might raise eyebrows. Probably won't be more than a few milliseconds, but we could end up with different minutes/hours/days if it occurs just as the clock rolls over.

Conch/lib/Conch/Orc/Lifecycle.pm
lseppala: Huh. This was on a different method. Thanks GH 👍

Conch/lib/Conch/Orc/Lifecycle.pm
lseppala: Resolved.

Conch/lib/Conch/Orc/Lifecycle.pm
lseppala: Resolved.

Conch/lib/Conch/Controller/Orc/Lifecycles.pm
```diff
@@ -0,0 +1,134 @@
...
+sub add_workflow ($c) {
+	return $c->status(403) unless $c->is_global_admin;
+
+	my $l = Conch::Orc::Lifecycle->from_id($c->param('id'));
+	return $c->status(404 => { error => "Not found" }) unless $l;
+
+	my $body = $c->validate_input('OrcLifecycleAddWorkflow') or return;
+	
+	my $w = Conch::Orc::Workflow->from_id($body->{workflow_id});
+	return $c->status(404 => { error => "Not found" }) unless $w;
```
lseppala: You might distinguish that the workflow ID doesn't exist, to avoid confusion that the URL is incorrect (goes for others in this file).

Conch/lib/Conch/Orc/Lifecycle.pm
sungo: @lseppala this comment :)

lseppala: Storing duplicates wasn't the issue. Sorry for the confusion. More, if you had two duplicates in `$self->plan` before calling `_rebuild_plan` (as is the case if an already-stored workflow is added in a request), it may silently alter or ignore the ordering. 

This is resolved by the checks you added in add_workflow and append_workflow. This means re-adding a previously stored workspace will always be ignored, even if the new request has a different order. One might consider returning an error in the controller if the workflow, but it's fine as it's consistent.

PR #226, opened 2018-04-06
feat: Endpoints for getting validation states with results

lseppala:
This adds two endpoints:

* `GET /device/:id/validation_state`
* `GET /workspace/:id/validation_state`

The first gets the most recent validation states (one per validation plan run) for a device. The second gets all most recent validations states for all devices in the workspace. The each validation state includes the associated validation results.

The purpose of this work is to supplant the legacy and problematic endpoint `GET /workspace/:id/problem` in `Conch::Controller::WorkspaceProblem`, which returns a tree structure of "failing" devices. The problem endpoint will be removed once device report ingestion is transitioned away from the legacy validation system.

### DIFF
--- a/Conch/lib/Conch/Model/ValidationState.pm
+++ b/Conch/lib/Conch/Model/ValidationState.pm
@@ -30,9 +30,9 @@ sub TO_JSON ($self) {
 		device_id          => $self->device_id,
 		validation_plan_id => $self->validation_plan_id,
 		status             => $self->status,
-		created            => Conch::Time->new( $self->created )->rfc3339,
+		created            => Conch::Time->new( $self->created ),
 		completed          => $self->completed
-			&& Conch::Time->new( $self->completed )->rfc3339
+			&& Conch::Time->new( $self->completed )
 	};
 }
 
--- a/Conch/lib/Conch/Model/ValidationState.pm
+++ b/Conch/lib/Conch/Model/ValidationState.pm
@@ -20,6 +20,22 @@ use constant {
 my $attrs = [qw( id device_id validation_plan_id status created completed )];
 has $attrs;
 
+=head2 TO_JSON
+
+=cut
+
+sub TO_JSON ($self) {
+	{
+		id                 => $self->id,
+		device_id          => $self->device_id,
+		validation_plan_id => $self->validation_plan_id,
+		status             => $self->status,
+		created            => Conch::Time->new( $self->created )->rfc3339,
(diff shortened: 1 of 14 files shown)


Conch/lib/Conch/Model/ValidationState.pm
```diff
@@ -20,6 +20,22 @@ use constant {
...
+
+sub TO_JSON ($self) {
+	{
+		id                 => $self->id,
+		device_id          => $self->device_id,
+		validation_plan_id => $self->validation_plan_id,
+		status             => $self->status,
+		created            => Conch::Time->new( $self->created )->rfc3339,
+		completed          => $self->completed
+			&& Conch::Time->new( $self->completed )->rfc3339
```
sungo: As of 901f47a95bbfa7a36a5af65f39cd29784a758c9b, you don't need to do the rfc3339 conversion yourself. Conch::Time has a TO_JSON method and if everyone just uses that, we can guarantee a consistent representation.

Conch/t/integration/05_v1_schema.t
```diff
@@ -31,10 +31,18 @@ my $uuid = Data::UUID->new;
...
+	description => 'Test Plan',
+	validations => [ { name => 'product_name', version => 1 } ]
+};
+
 my $t = Test::MojoSchema->new(
 	Conch => {
 		pg      => $pgtmp->uri,
-		secrets => ["********"]
+		secrets => ["********"],
+		features => { new_validation => 1 },
```
sungo: Is there actually a feature flag for the new validation system? There's one in conch.conf.dist but I can't find where it is used. I think it'd probably be a good idea to feature flag it, if we can, so that we can switch to the old system without a full rollback.

Conch/lib/Conch/Model/ValidationState.pm
lseppala: Cool! Will remove

Conch/t/integration/05_v1_schema.t
lseppala: It is. It's used within the DeviceReport controller. See https://github.com/joyent/conch/blob/master/Conch/lib/Conch/Controller/DeviceReport.pm#L85-L90

sungo:
To c&p what I said in chat, one of the test runs failed. Which is weird since the other succeeded on the same git rev. I saw this end of last week too and rerunning the build cleared the error.  I have no idea why we'd see a transient failure in the tests.

PR #229, opened 2018-04-12
fix: UI Login improvements

lseppala:
* Don't throw incorrect credential errors to Rollbar
* Display message when wrong user/password
* Clear the fields on login, just clear password field on bad login


lseppala:
Split the API changes into https://github.com/joyent/conch/pull/232

PR #236, opened 2018-04-18
refactor: simplify Validation schema

lseppala:
This simplifies the validation schema by making all top-level keys in the schema 'required' when creating the JSON schema for validation.

I had this idea while writing the validation tutorial. Makes it a bit more simple.


sungo:
Am I understanding the patch correctly and it is no longer possible to have optional top-level data elements?

lseppala:
I made a changed discussed in MM. By default, all top-level attributes are required by default, but you may specify a `required` attribute to override and specify exactly which attributes are required.

PR #240, opened 2018-04-26
Add ability to update orc lifecycles and some bug fixes

sungo:

### DIFF
--- a/Conch/lib/Conch/Controller/Orc/Lifecycles.pm
+++ b/Conch/lib/Conch/Controller/Orc/Lifecycles.pm
@@ -56,6 +56,13 @@ sub create ($c) {
 		});
 	}
 
+	if(Conch::Orc::Lifecycle->from_role_id($body->{role_id})) {
+		return $c->status_with_validation(400, Error => {
+			error => "Role is already in use"
+		});
+	}
+
+
 	$body->{version} = 0 unless $body->{version};
 	if(Conch::Orc::Lifecycle->from_name($body->{name})) {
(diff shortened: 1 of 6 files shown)


Conch/lib/Conch/Controller/Orc/Lifecycles.pm
```diff
@@ -62,6 +62,30 @@ sub create ($c) {
 	$c->status(303 => "/o/lifecycle/".$l->id);
 }
 
+=head2 update
+
+Update an existing lifecycle
+
+=cut
+
+sub update ($c) {
```
lseppala: We should be checking a role ID in the payload is valid and exists in the database if present.

lseppala: Also need to check if the role ID is not already taken by another lifecycle, since it's unique.

Conch/lib/Conch/Controller/Orc/Lifecycles.pm
```diff
@@ -62,6 +75,45 @@ sub create ($c) {
...
+
+		my $lr = Conch::Orc::Lifecycle->from_role_id($body->{role_id});
+		if ($lr and ($lr->id ne $l->id)) {
+			return $c->status_with_validation(400, Error => {
+				error => "Role is already in use"
+			});
+		}
+	}
+
+	$body->{version} = 0 unless $body->{version};
```
lseppala: version is no longer in the schema. Also on line 66

PR #243, opened 2018-04-30
feat: add query parameter filtering for validation states

lseppala:
Provide query parameters to limit the validation states by status  ('pass', 'fail', 'error'). This is to improve UI and CLI responsiveness, as users care most about the validation states that failed, e.g., the 'Problems' page.

### DIFF
--- a/Conch/lib/Conch/Controller/DeviceValidation.pm
+++ b/Conch/lib/Conch/Controller/DeviceValidation.pm
@@ -13,6 +13,7 @@ package Conch::Controller::DeviceValidation;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use Conch::Models;
+use List::Util qw(notall any);
 
 =head2 list_validation_states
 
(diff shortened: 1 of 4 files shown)


Conch/lib/Conch/Controller/DeviceValidation.pm
```diff
@@ -21,11 +22,33 @@ Get latest validation states for a device
 =cut
 
 sub list_validation_states ($c) {
+	my @statuses;
```
sungo: perhaps this should use the new json schema based validation for api input?

Conch/lib/Conch/Controller/DeviceValidation.pm
```diff
@@ -21,11 +22,33 @@ Get latest validation states for a device
...
+	if (
+		@statuses
+		&& notall {
+			my $a = $_;
+			any { $_ eq $a } qw( pass fail error )
+		}
+		@statuses
+		)
+	{
+		return $c->status(
```
sungo: Is there no way now to get all validation states? I'd like to retain that ability while also supporting this kind of filtering.

Conch/lib/Conch/Controller/DeviceValidation.pm
lseppala: 🤦‍♂️ 

duh.

Conch/lib/Conch/Controller/DeviceValidation.pm
lseppala: If unspecified, all validation states are returned. Corresponding test https://github.com/joyent/conch/pull/243/files#diff-86a28648deb62650c8e4a8eb6d5cf155R461

Conch/lib/Conch/Controller/DeviceValidation.pm
lseppala: This is validating the values in query parameters, not in a JSON payload. We can still use JSON::Validator to validate the parameters. Here's what it looks like:

```
	my @statuses;
	if ( $c->param('status') ) {
		@statuses = map { lc($_) } split /,\s*/, $c->param('status');
		my @errors = validate_json(
			\@statuses,
			{
				type  => 'array',
				items => { type => 'string', enum => [ 'pass', 'fail', 'error' ] }
			}
		);
		return $c->status(
			400,
			{
				error =>
					"'status' query parameter must be any of 'pass', 'fail', or 'error'."
			}
		) if @errors;
	}
```

It's a little less obtuse than the previous code, and I like having a standard way of validating any input. Thoughts?

PR #246, opened 2018-05-03
feat: Add JWT as a method of authentication

lseppala:
Add JSON Web Tokens (JWT) as an alternative method of authentication. This change is backwards compatible with previous authentication methods.

I chose to add this rather than creating loop-holes for splitting out UI development into a separate repo. This is what we all want anyways, so why not?

Here's a couple of important points:

* Cookie and Basic auth left as-is. They can be removed later if desired.
* JWT are canonically encoded in Base64 and returned in the JSON response of a successful login in the `token` attribute. Authentication with a JWT is made with the `Authentication: Bearer <token>` header.
* Like cookies, we use the `secrets` in the configuration file to provide the secure HMAC signature. Similarly, secrets can be rotated and JWTs with old but still available secrets will work. Changing all secrets invalidates all JWTs.
* JWTs do not expire.

I expect discussion about the last point, and held off on implementing anything before discussing. Here are some points:

* Non-expiring JWTs mean the CLI doesn't need to be re-profiled after 30 days (desirable?)
* We can expire tokens in client implementation. For example, in the UI, we can drop a stored token after _x_ minutes/hours of inactivity.
* If we decide to add expiration, `GET /login` with a valid token can re-issue a new token
* Revoking all tokens can be done by changing the configured `secrets`. 

Let me know your thoughts.

### DIFF
--- a/Conch/lib/Conch/Model/SessionToken.pm
+++ b/Conch/lib/Conch/Model/SessionToken.pm
@@ -53,7 +53,8 @@ Check if a session token is valid for the user ID. Returns 1 if valid, 0 otherwi
 
 sub check_token ( $class, $user_id, $token ) {
 
-	Conch::Pg->new->db->query( 'select delete_expired_tokens()' );
+	Conch::Pg->new->db->delete( 'user_session_token',
+		{ expires => { '<=' => 'now()' } } );
 
 	return Conch::Pg->new->db->query(
 		q{
@@ -68,7 +69,6 @@ sub check_token ( $class, $user_id, $token ) {
 	)->rows;
 }
 
-
 =head2 use_token
 
 Use a token by permanetly deleting it from the database. Will return 1 if the
@@ -78,7 +78,8 @@ token was present and valid, 0 otherwise.
 
 sub use_token ( $class, $user_id, $token ) {
 
-	Conch::Pg->new->db->query( 'select delete_expired_tokens()' );
+	Conch::Pg->new->db->delete( 'user_session_token',
+		{ expires => { '<=' => 'now()' } } );
 
 	return Conch::Pg->new->db->query(
 		q{
--- a/Conch/lib/Conch/Model/SessionToken.pm
+++ b/Conch/lib/Conch/Model/SessionToken.pm
@@ -1,0 +1,120 @@
+=pod
+
+=head1 NAME
+
+Conch::Model::SessionToken
+
+=head1 METHODS
+
+=cut
+
+package Conch::Model::SessionToken;
+use Mojo::Base -base, -signatures;
+
+use Conch::UUID qw(is_uuid);
+use Session::Token;
+
+has [
+	qw(
+		email
+		id
+		name
+		password_hash
+		)
+];
+
+=head2 create
+
+Create and record a session token
+
+=cut
+
+sub create ( $class, $user_id, $expires ) {
+	my $token = Session::Token->new->get;
+
+	Conch::Pg->new->db->query(
+		q{
+			insert into user_session_token
+				(user_id, token_hash, expires)
+			values( ?, digest(?, 'sha256'), to_timestamp(?)::timestamptz)
+		},
+		$user_id,
+		$token,
+		$expires,
+	);
+	return $token;
+}
+
+=head2 check_token
+
+Check if a session token is valid for the user ID. Returns 1 if valid, 0 otherwise
+
+=cut
+
+sub check_token ( $class, $user_id, $token ) {
+
+	Conch::Pg->new->db->query( 'select delete_expired_tokens()' );
+
+	return Conch::Pg->new->db->query(
+		q{
+			select 1
+			from user_session_token
+			where
+				user_id = ? and
+				token_hash = digest(?, 'sha256')
+		},
+		$user_id,
+		$token
+	)->rows;
+}
+
+
+=head2 use_token
+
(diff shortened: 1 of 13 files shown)


lseppala:
I broke down and added a expiry check using the same timeout as the cookie sessions. If we decide to do something different, I can drop the commit.

lseppala:
I read [an interesting authorization implementation with JWT]( https://medium.com/lightrail/getting-token-authentication-right-in-a-stateless-single-page-application-57d0c6474e3) and have thought about how it might apply to our use.

JWT are encoded as a base64 string with the format `header.payload.signature`. When a user logs in, the JWT is split into two parts in the response: the `header.payload` is returned in the response body, and the `signature` is returned as an HttpOnly cookie. Javascript in the browser will only ever have access to the `header.payload` part. Javascript can pull out any user information from the `payload` section that may be useful in the interface,  it can add the token to an `Authentication` header for requests. Meanwhile, the browser will send the `HttpOnly` cookie with every request as long as the cookie hasn't expired. The API will require both the `header.payload` and the `signature` parts, and will combine them to verify authorization.

Non-browser clients like Conch shell can also use this scheme. Unlike Javascript in the browser, a client can directly access the HttpOnly cookie value. Clients can then combine the two parts themselves and return the full `header.payload.signature` token in the Authorization header. This allows them to by-pass needing to store or send the cookie for every request.

By splitting the JWT into a cookie and a response body, we gain the following advantages:

1. It completely disallows CSRF attacks, as the `header.payload` part in the response will be inaccessible with properly configured CORS headers.
2. It limits XSS attacks, as scripts won't have direct access to the full token (it could still submit requests on behalf of the user while viewing our page, however)
3. It allows browser clients and non-browser clients to have two different session expirations. For example, the HttpOnly cookie can be set to expire after 30 minutes. The browser won't send the cookie after it has expired and the user will be logged out. This will limit any attacks where a user left a browser session open.
4. Meanwhile, a separate expiration may be set for the JWT as a whole for a much longer time period (say, a month). This allows non-browser clients like Conch-shell to have much longer sessions while still using the same scheme as the browser.

The big disadvantage is the complexity this introduces for clients. In pseudo-Perl, authorizing a request would look like the following

```
$cookie_jar = set_up_cookie_jar()
$response = http->post("https://conch.joyent.us/login", json => $credentials)
$jwt_signature = cookie_jar->get("jwt-signature")
$jwt_payload = response->get("jwt-payload")
$jwt_token = "$jwt_signature.jwt_payload"

$authorized_response = http->get("http://conch.joyent.us/some_authorized_resource", Authorization => $jwt_token)
```

Unfortunately, there's no perfect scheme, and each involves trade-offs between security and user friendliness. I like this scheme because it provides a single scheme for both browser and non-browser clients, it allows for the different expiration periods, and it addresses a host of security issues. I'd like to hear any thoughts and feedback.

sungo:
To carry over conversations held elsewhere, I am a fan of the more complex scheme, specifically because it would allow us to special-case particular clients. For the mainline clients, I'm not worried about the added complexity because we'll implement it once for the shell (probably in go-conch so any go client will benefit) and once in the web UI.

My bigger concern is the unofficial clients where folks are using httpie or curl. If anyone is still copying conch cookies around, that'll obviously break. HTTP basic auth users wouldn't break, I don't think.

I'd also like to have a conversation elsewhere about HTTP basic auth and our ongoing support for it.

Conch/lib/Conch/Model/SessionToken.pm
```diff
@@ -0,0 +1,120 @@
...
+
+=head2 check_token
+
+Check if a session token is valid for the user ID. Returns 1 if valid, 0 otherwise
+
+=cut
+
+sub check_token ( $class, $user_id, $token ) {
+
+	Conch::Pg->new->db->query( 'select delete_expired_tokens()' );
```
sungo: Can we make this a method or something that has a descriptive name? While the sql is clear, I had to go to the sql to verify that this was acting on the daatbase and wasn't just a select that we were throwing away.

sql/migrations/0030-user-session-token.sql
```diff
@@ -0,0 +1,18 @@
...
+		user_id    uuid        not null references user_account(id),
+		token_hash bytea       not null,
+		expires    timestamptz not null,
+		primary key (user_id, token_hash)
+	);
+
+	-- fast deleting of expired otkens
+	create index on user_session_token (expires asc);
+
+	create function delete_expired_tokens() returns void
```
sungo: Could this maybe return the number of tokens that were deleted?

Conch/lib/Conch/Model/SessionToken.pm
lseppala: We've been talking about stored procedures, and I saw this as a way to lightly try them out (originally, I was going to use this function in a trigger to force ensure you could never query for expired tokens, but then I realized that automatically disappearing data wasn't the best idea...).

This disconnect between the code and the query is the biggest reason I am hesitant to use stored procedures. It certainly has its use in more complex queries (like the `workspace_devices` procedure I made yesterday), but is probably not appropriate here. I'd prefer to go ahead and replace the invocation with the actual queries.

sungo: You're reusing the query so either way it should be abstracted a bit. I have no problem with the stored procedure itself but it just needs to be a bit more clear in the code, via either a good method name or docs or both.

PR #251, opened 2018-05-10
BREAKING: remove orchestration

sungo:
We've decided to reevaluate the orchestration system from first principles. Since we won't be using this code for the foreseeable future, if ever, remove it from the code base to avoid bitrot.


sql/migrations/0030-remove-orc.sql
```diff
@@ -0,0 +1,15 @@
+SELECT run_migration(30, $$
```
lseppala: 30 taken by #246

lseppala: Next avail is 32 (31 was taken by #249 )

sungo: Updated to 32

PR #263, opened 2018-05-17
Remove v1 references

### LINKED ISSUE #255: Remove `v1` references
We've decided a while ago that we do not want to version the API, and instead using client version checks to check for compatibility. There are many references to `v1` in the code base currently. Before cutting to the `v3` release version, let's cleans these out by renaming and removing any `v1` terminology.

Instances:
* `json-schema/v1.json`
* `as_v1_json`, `as_v1` in Perl code

lseppala:
We've firmly decided against versioning the API for now (instead relying on client-side version verification). The myriad references to `v1` are therefore confusing and useless. This PR renames or removes the references.

Closes #255.


Conch/lib/Conch/Controller/DeviceLocation.pm
```diff
@@ -29,7 +29,7 @@ sub get ($c) {
 		{ error => "Device $device_id is not assigned to a rack" } )
 		unless $maybe_location;
 
-	$c->status( 200, $maybe_location->as_v1_json );
+	$c->status( 200, $maybe_location->TO_JSON );
```
sungo: You shouldn't need that direct call. Mojo will call TO_JSON for you when it's appropriate.

Conch/lib/Conch/Controller/HardwareProduct.pm
```diff
@@ -24,7 +24,7 @@ Class::HardwareProduct objects
 
 sub list ($c) {
 	my $hardware_products = Conch::Model::HardwareProduct->new->list;
-	$c->status( 200, [ map { $_->as_v1_json } @$hardware_products ] );
+	$c->status( 200, [ map { $_->TO_JSON } @$hardware_products ] );
```
sungo: Shouldn't need that direct call to TO_JSON. Mojo should handle that for you.

Conch/lib/Conch/Controller/DeviceLocation.pm
lseppala: Yup. I did a global find-and-replace for the string and intentionally didn't do any functional refactors like this. Do you want me to include this change in this PR, or have a separate?

sungo: Let's go ahead and include it in this PR and just do the transition all at once

PR #267, opened 2018-06-06
CRUD for datacenter_rack_role and datacenter_rack

sungo:


Conch/lib/Conch/Controller/DatacenterRack.pm
```diff
@@ -0,0 +1,146 @@
...
+
+}
+
+=head2 get
+
+Get a single rack
+
+=cut
+
+sub get ($c) {
```
lseppala: For both get and get_all, should restrict to global admin.

You can also pull this logic into `under`, which allows you to de-dup the check for update and delete

Conch/lib/Conch/Controller/DatacenterRackRole.pm
```diff
@@ -0,0 +1,127 @@
...
+
+}
+
+=head2 get
+
+Get a single rack role
+
+=cut
+
+sub get ($c) {
```
lseppala: Likewise, should restrict to global admin I think.

Conch/lib/Conch/Controller/DatacenterRack.pm
sungo: Fixed.

Conch/lib/Conch/Controller/DatacenterRackRole.pm
sungo: Fixed.

Conch/lib/Conch/Controller/DatacenterRack.pm
```diff
@@ -0,0 +1,154 @@
...
+	$c->status(200, Conch::Model::DatacenterRack->all());
+}
+
+
+=head2 update
+
+=cut
+
+sub update ($c) {
+	return $c->status(403) unless $c->is_global_admin;
```
lseppala: Dup from `under` logic. Same with `delete`

PR #279, opened 2018-06-22
DB migration to add hardware product spec and SKU fields

bdha:
We are changing how we manage product names and SKUs internally.

We would also like a more expressive way of defining hardware product specifications (currently stored in the hardware_product_profile table.)

This migration adds a `sku` column, and changes the current `name` column to `product_name`. So we end up with:

* name: manta-storage-v3-512g-36-8tb (Conch specific)
* sku: 600-0025-001
* product_name: Joyent-S10G4

I think we agreed to not manage table contents in migrations, so I'll update the fields after deployment.

We will also need to update DB/API code to ref the new columns.

An example of what we end up with:

```
conch=# select name,alias,sku,product_name from hardware_product where sku is not null;
             name              |         alias         |     sku      | product_name 
-------------------------------+-----------------------+--------------+--------------
 manta-storage-v3-512g-36-8tb  | Mantis Shrimp MkIII   | 600-0025-001 | Joyent-S10G4
 manta-storage-v3-256g-36-10tb | Mantis Shrimp MkIII   | 600-0025-002 | Joyent-S10G4
 manta-storage-v3-256g-36-12tb | Mantis Shrimp MkIII.5 | 600-0036-001 | Joyent-S10G5
(3 rows)
```

### DIFF
--- a/Conch/docs/hw-specs/manta-storage-v3-256g-36-10tb.json
+++ b/Conch/docs/hw-specs/manta-storage-v3-256g-36-10tb.json
@@ -1,0 +1,103 @@
+{
+	"chassis": {
+		"dmi": {
+			"bios": {
+				"vendor_name": "American Megatrends Inc.",
+				"version": "3.0a"
+			},
+			"bmc": {
+				"version": "3.65"
+			},
+			"system": {
+				"vendor_id": "66fed141-85c1-4136-80d7-b1fffbfd2991",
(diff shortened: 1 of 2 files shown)


daleghent:
LGTM

sungo:
This PR did not include updates to the perl libraries or API endpoints that deal with hardware products. Going forward, please include code updates that match the database changes.

PR #286, opened 2018-07-09
Fix hardware product lookups.

bdha:
Due to the recent changes in hardware product naming, we need to modify how we do validations based on the device type.

If we're a switch, we continue on as we have been.
If we're a server, we need to use the SKU or the generation_name depending on what kind of data the reporter is sending up.
In the future, we'll want to further normalize this -- or add multiple validations for each field. This is a bit of a crunch hack, however.

### DIFF
--- a/Conch/lib/Conch/Controller/DeviceReport.pm
+++ b/Conch/lib/Conch/Controller/DeviceReport.pm
@@ -26,8 +26,18 @@ sub process ($c) {
 	my $raw_report = $c->req->body;
 
 	my $hw_product_name = $device_report->{product_name};
-	my $maybe_hw =
-		Conch::Model::HardwareProduct->lookup_by_name($hw_product_name);
+	my $maybe_hw;
+
+	if ( $device_report->{device_type}
+		&& $device_report->{device_type} eq "switch" )
+	{
+		$maybe_hw =
+			Conch::Model::HardwareProduct->lookup_by_name($hw_product_name);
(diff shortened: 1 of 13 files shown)


Conch/lib/Conch/Controller/DeviceReport.pm
```diff
@@ -26,8 +26,18 @@ sub process ($c) {
 	my $raw_report = $c->req->body;
 
 	my $hw_product_name = $device_report->{product_name};
-	my $maybe_hw =
-		Conch::Model::HardwareProduct->lookup_by_name($hw_product_name);
+	my $maybe_hw;
+
+	if ( $device_report->{device_type}
+		&& $device_report->{device_type} eq "switch" )
```
daleghent: For imminent future-proofing this test, we might want to lop PDUs (`pdu`) in with this as well.

Conch/lib/Conch/Legacy/Control/DeviceReport.pm
```diff
@@ -45,12 +45,22 @@ Record device report and device details from the report
 =cut
 sub record_device_report {
 	my ( $schema, $dr, $raw_report ) = @_;
-	my $hw = $schema->resultset('HardwareProduct')->find(
+	if ($dr->{device_type} && $dr->{device_type} eq "switch")
```
daleghent: PDU consideration here as well

PR #287, opened 2018-07-09
fix: Update switchpeers validation to support Arista deployments

bdha:
We need to detect the peer vendor so we can modify the peer text (Arista hands us "Ethernet"). We've also modified the map so data is ports 1 - 20 and admin is 24 - 44.

### DIFF
--- a/Conch/lib/Conch/Validation/SwitchPeers.pm
+++ b/Conch/lib/Conch/Validation/SwitchPeers.pm
@@ -87,24 +87,32 @@ sub validate {
 
 sub _calculate_switch_peer_ports {
 	my ( $self, $rack_unit, $rack_slots, $peer_vendor ) = @_;
-	my $rack_index =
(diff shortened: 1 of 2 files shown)


Conch/lib/Conch/Validation/SwitchPeers.pm
```diff
@@ -26,14 +26,21 @@ sub validate {
...
 
-	my @peer_ports =
-		$self->_calculate_switch_peer_ports( $device_location->rack_unit,
-		$rack_slots );
-
 	my @eth_nics =
 		map { $data->{interfaces}->{$_} }
 		grep { $_ =~ /eth/ } ( keys $data->{interfaces}->%* );
 
+	# XXX This is awful. I'm sorry.
```
sungo: Let's find a way to do this without another "this is awful" comment in the codebase. I'll take a look.

karenetheridge: what is awful about this?

PR #291, opened 2018-07-12
refactor: the unparsable logger is no longer used with the new validation system so remove the dead code

sungo:


bdha:
The only good code is deleted code. +1

PR #296, opened 2018-07-16
In audit logs, don't log JWT data and log only json in each line

sungo:


Conch/lib/Conch.pm
```diff
@@ -210,7 +210,11 @@ sub startup {
 		}
 
 		my $log_path = $opts{log_path} || "log/audit.log";
-		my $log = Mojo::Log->new(path => $log_path);
+		my $log = Mojo::Log->new(path => $log_path, short => 1);
+		$log->format(sub {
+			my ($time, $level, @lines) = @_;
+			return Mojo::JSON::to_json($lines[0])."\n";
```
perigrin: Don't we call this with a non-ref at `lib/Conch/ValidationSystem.pm#L68`? Quick tests say this shouldn't be a problem but if Mojo::JSON ever changes to be more like JSON::XS about it ...

sungo: In theory, and testing, the one created here in Audit is an independent logger. That's probably a problem all its own since we can't configure the logger outside the code.

karenetheridge: to_json should probably be encode_json - IMO we should be able to handle non-ascii data everywhere and DTRT.

sungo: Done

PR #298, opened 2018-07-17
when handling an exception, use the exception that is always in the stash but not always in the args

sungo:
This fixes the scenario where a request just times out if DBIC throws an exception

### DIFF
--- a/Conch/lib/Conch.pm
+++ b/Conch/lib/Conch.pm
@@ -147,7 +147,7 @@ sub startup {
 			my ( $c, $args ) = @_;
 			return unless my $template = $args->{template};
 			if ( $template =~ /exception/ ) {
-				my $exception = $c->stash('exception');
+				my $exception = $c->stash('exception') // $args->{exception};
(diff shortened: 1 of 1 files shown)


Conch/lib/Conch.pm
```diff
@@ -147,7 +147,7 @@ sub startup {
 			my ( $c, $args ) = @_;
 			return unless my $template = $args->{template};
 			if ( $template =~ /exception/ ) {
-				my $exception = $args->{exception};
+				my $exception = $c->stash('exception');
```
karenetheridge: Is it *always* in the stash? i.e. we don't have to do:

    my $exception = $c->stash('exception') // $args->{exception};

?

karenetheridge: I'll approve this now so you can merge whenever.

sungo: In my testing, yes, it's always in the stash. I can add that safety valve if you'd like though

PR #304, opened 2018-07-20
command line utility to create a user, optionally sending an email

karenetheridge:
for https://app.liquidplanner.com/space/174715/projects/show/46723720

This kind of happened accidentally as I was working through the user db code and discovered there wasn't an easy way of creating a user -- indeed it seems the only way this happens is through the workspace controller (!)

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -37,6 +37,7 @@ requires 'Types::UUID';
 requires 'Role::Tiny';
 requires 'List::MoreUtils';
 requires 'List::MoreUtils::XS';
+requires 'Getopt::Long::Descriptive';
 
 ### Legacy Deps
 #
--- a/cpanfile.snapshot
+++ b/cpanfile.snapshot
@@ -1639,6 +1639,25 @@ DISTRIBUTIONS
       perl 5.014000
       strict 0
       warnings 0
+  Getopt-Long-Descriptive-0.102
+    pathname: R/RJ/RJBS/Getopt-Long-Descriptive-0.102.tar.gz
+    provides:
+      Getopt::Long::Descriptive 0.102
+      Getopt::Long::Descriptive::Opts 0.102
+      Getopt::Long::Descriptive::Usage 0.102
+    requirements:
+      Carp 0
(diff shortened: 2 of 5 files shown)


karenetheridge:
let's wait to merge this until after the Conch subdir is nuked, and then I will rebase.

Conch/bin/conch-create-user
```diff
@@ -0,0 +1,71 @@
...
+    [ 'password|p=s',   'the user password', { default => random_password() } ],
+    [ 'send-email',     'send email to user' ],
+    [],
+    [ 'help',           'print usage message and exit', { shortcircuit => 1 } ],
+);
+
+print($usage->text), exit if $opt->help;
+
+my $conch = Mojo::Server->new->build_app('Conch');
+my $schema = Conch::DB::Schema->connect($conch->config('pg'));
```
sungo: if you're loading up the Conch mojo app, there's a 'schema' helper available that does the needful. https://github.com/joyent/conch/blob/master/Conch/lib/Conch.pm#L54

Unless I'm misunderstanding which is always possible :)

karenetheridge: that's a Mojo schema object, relies on the Mojo way of connecting, which is the bits I'm aiming to remove. That's why I added new connect logic right in Conch::DB::Schema (likely soon to be renamed Conch::DB, although it `isa->('DBIx::Class::Schema')`).

karenetheridge: The reason why I'm loading up the app there is to get at the config tree.. there's probably a better/faster way of doing that, so I'll replace that when I find it.

sungo: that helper is returning a Conch::DB::Schema object configured using the data in the mojo config.

karenetheridge:
I'm going to clean up the db handling a bit more, and move this to the Conch::Command:: namespace (now that I know about Mojolicious::Command::*)

karenetheridge:
this PR contains (as it depends upon) #306.

karenetheridge:
this is ready to merge.

PR #306, opened 2018-07-20
move Conch::DB::Schema -> Conch::DB; remove moose

karenetheridge:
This also contains a small change to a Makefile rule, and lots of fluff to normalize formating.


karenetheridge:
this is good to merge.

PR #311, opened 2018-07-26
db refactoring and change password interface

karenetheridge:
- more DBIC usage - Conch::Model::User and Conch::Model::SessionToken are now fully converted
- db_* helper methods added, for easy access to db queries
- LP-46272723: user can change their own password


lib/Conch/Controller/User.pm
```diff
@@ -134,16 +149,19 @@ Get the individual key/value pair for a setting for the User
...
+		{ name => $key },
+		{ order_by => { -desc => 'created' } },
+	)->first;
 
 	return $c->status( 404, { error => "No such setting '$key'" } )
-		unless $settings->{$key};
+		unless $setting;
 
-	$c->status( 200, { $key => $settings->{$key} } );
+	$c->status( 200, { $key => from_json($setting->value) } );
```
sungo: the 'status' helper should be handling the serialization for you at https://github.com/joyent/conch/blob/master/lib/Conch.pm#L109

*should* is operative, I think.

lib/Conch/Controller/User.pm
```diff
@@ -155,21 +173,48 @@ Delete a single setting for a user, provided it was set previously
...
+Stores a new password for the current user.
+
+Optionally takes a query parameter 'clear_tokens' (defaulting to true), to also revoke session
+tokens for the user, forcing all tools to log in again.
+
+=cut
+
+sub change_password ($c) {
+	my $body = $c->req->json;
+	return $c->status( 400, { error => 'Payload required' } ) unless $body;
```
sungo: Let's move this to use the json validator like https://github.com/joyent/conch/blob/master/lib/Conch/Controller/DB/HardwareProduct.pm#L84

lib/Conch/Controller/User.pm
karenetheridge: this is deserialization, from the db.

sungo: Roger

lib/Conch/Controller/User.pm
karenetheridge:
```
    #          $got->{TEST} = '"TEST"'
    #     $expected->{TEST} = 'TEST'
```

karenetheridge: ok, I'm adding an entry to json-schema/input.yaml, and then this to `change_password`:

```
 sub change_password ($c) {
-       my $body = $c->req->json;
-       return $c->status( 400, { error => 'Payload required' } ) unless $body;
+       my $body =  $c->validate_input('UserPassword');
+       return if not $body;
 
        my $new_password = $body->{password};
-       return $c->status(400, { error => 'Missing parameter "password"' }) unless $new_password;
 
```

PR #315, opened 2018-07-30
fix an issue in the status helper where we were operating on main app and not the controller

sungo:


karenetheridge:
This PR doesn't do what you think it does -- but it's still fine.  The `$self` you changed to `$c` was the controller all along -- it's what called the `status` sub, which was a controller. So what you really did was just change a variable name :)

sungo:
Not according to my log output. When setting up the new logger, I was tracking an issue where `status()` was recording a different request id than the rest of the request path. DDP'ing `$self` there yielded the app, not the controller.

karenetheridge:
The real wtf (IMHO) in the status sub is that the app is being referenced as `$c`, when it really should be `$app`. `$c` should only be used to reference a controller, lest everyone get totally confused (which I suspect happened to most people driving through here.)

FWIW I have a commit in my long-running-and-not-merged branch which changed these variable names, but I haven't added it to a PR because I didn't want to muck around with the helpers just yet.

karenetheridge:
Interesting. What was the call path?  When does something call `->status` when that thing isn't a controller?

lib/Conch.pm
```diff
@@ -96,12 +95,12 @@ sub startup {
 
 			if ($code >= 400) {
 				$msg = "$msg || Payload: ".Mojo::JSON::to_json($payload);
-				$c->log->warn($msg);
+				$c->app->log->warn($msg);
```
karenetheridge: This line has an operational difference.
And now, anytime `$self` appears in the helper, it is the original app, via a closure. (This would be a memory leak if the status sub was ever generated more than once, which I don't think it is in mojo.)

sungo:
Apologies. I'd forgotten a piece of the diff.

https://github.com/joyent/conch/pull/315/files#diff-b54afa5d746afe7a2bdc214ee0415388L82

`$c` was an app because the helper specifically assigned it for some reason.

lib/Conch.pm
sungo: That'll get cleaned up in the PR coming from sungo/moar-logging

PR #318, opened 2018-08-01
features: new /user interfaces for adding, deleting users and resetting password

karenetheridge:
features: new /user interfaces for adding, deleting users and resetting password
    
        DELETE  /user/<id>/password
        POST    /user          json payload: { name: ..., email: ..., password: <optional> }
        DELETE  /user/<id>
    
    This change includes a database migration.

    fulfills
        - LP-46272834 (forced password reset)
        - LP-46519267 (deactivate users rather than deleting them)

### DIFF
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -40,9 +40,6 @@ Revoke a specified user's session tokens. Global admin only.
 =cut
 
 sub revoke_user_tokens ($c) {
-	return $c->status( 403, { error => 'Must be global admin' } )
-		unless $c->is_global_admin;
-
 	my $user_param = $c->stash('target_user');
 	my $user =
 		is_uuid($user_param) ? $c->db_user_accounts->lookup_by_id($user_param)
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -185,7 +185,7 @@ sub delete_setting ($c) {
 	return $c->status(204);
 }
 
-=head2 change_password
+=head2 change_own_password
 
 Stores a new password for the current user.
 
@@ -194,7 +194,7 @@ tokens for the user, forcing all tools to log in again.
 
 =cut
 
-sub change_password ($c) {
+sub change_own_password ($c) {
 	my $body =  $c->validate_input('UserPassword');
 	return if not $body;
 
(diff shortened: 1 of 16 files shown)


lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,127 @@ sub change_password ($c) {
...
+	}
+
+	if ($c->req->query_params->param('clear_tokens') // 1) {
+		$user->delete_related('user_session_tokens');
+
+		# TODO: set a flag in the user db to signal that the session token should not be
+		# accepted?
+	}
+
+	return $c->status(204);
```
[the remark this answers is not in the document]
perigrin: Well it should be a 204 if `send_password_reset_mail` is false because then we've finished all the processing we're going to do.

lib/Conch/Route.pm
```diff
@@ -37,34 +37,49 @@ Set up the full route structure
...
-	my $unsecured = shift;	# this is the base routing object
+	my $root = shift;	# this is the base routing object
 	my $features = shift || {};
 
+	# provides a route to chain to that first checks the user is a global admin.
+	$root->add_shortcut(require_global_admin => sub {
+		my ($r, $path) = @_;
+		$r->any(sub {
+			my $c = shift;
+			return $c->status( 403, { error => 'Login required' } )
```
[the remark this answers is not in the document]
karenetheridge: does conch-shell or conch-api rely on 403 vs 401? (this change is just a translation of what was already in Conch::Controller::User::revoke_user_tokens.)

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,127 @@ sub change_password ($c) {
...
+=cut
+
+sub reset_user_password ($c) {
+	my $user_param = $c->stash('target_user');
+	my $user =
+		is_uuid($user_param) ? $c->db_user_accounts->lookup_by_id($user_param)
+	  : $user_param =~ s/^email\=// ? $c->db_user_accounts->lookup_by_email($user_param)
+	  : undef;
+
+	return $c->status( 404, { error => "user $user_param not found" } )
```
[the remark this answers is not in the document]
karenetheridge: should we change that now, or overhaul the responses in a later change (and cover the entire app when we do so)? I'm leaning towards the latter.

karenetheridge:
ok, I can add more logging :)

lib/Conch/Route.pm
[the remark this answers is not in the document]
sungo: The shell has a helper to handle http responses: https://github.com/joyent/conch-shell/blob/master/pkg/conch/errors.go#L63

A change from a 403 to a 401 will change the response value to a different error but it shouldn't be an issue. The only substantive change will be to the text the user sees, since all the functions bail when any error occurs.

In this specific case, I think an "unauthorized" message will be more descriptive and clear than "forbidden".

karenetheridge:
Changes pushed; ready for re-review.

To see the changes, `git diff 36054ba3 6b21281f`

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,136 @@ sub change_password ($c) {
...
+	  : undef;
+
+	# user not found, but secure applications don't let on to the user
+	if (not $user) {
+		$c->log->debug("user lookup for $user_param failed");
+		return $c->status(202);
+	}
+
+	my $new_password = $c->random_string();
+	$c->log->debug('user ' . $c->stash('user')->name . ' resetting password for user ' . $user->name);
```
sungo: nit: let's bump that up to warn or some such.

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,136 @@ sub change_password ($c) {
...
+	$user->update({ password => $new_password });
+
+	if ($c->req->query_params->param('clear_tokens') // 1) {
+		$c->log->debug('user ' . $c->stash('user')->name . ' deleting user session tokens for for user ' . $user->name);
```
sungo: nit: bump to warn

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,136 @@ sub change_password ($c) {
...
+Creates a user. Global admin only.
+
+Optionally takes a query parameter 'send_invite_mail' (defaulting to true), to send an email
+to the user with the new password.
+
+=cut
+
+sub create ($c) {
+	my $body =  $c->validate_input('NewUser');
+	return if not $body;
```
sungo: warn about this please

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,136 @@ sub change_password ($c) {
...
+
+	# this would cause horrible clashes with our /user routes!
+	return $c->status(400, { error => 'user name "me" is prohibited', }) if $name eq 'me';
+
+	# we don't use lookup_by_* because they only search active users.
+	if (my $user = $c->db_user_accounts->search({
+			-or => { name => $name, email => $email }
+		})->first)
+	{
+		return $c->status(409, {
```
sungo: log about this please

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,136 @@ sub change_password ($c) {
...
+	}
+
+	if ($user->deactivated) {
+		return $c->status(410, {
+			error => 'user was already deactivated',
+			user => { map { $_ => $user->$_ } qw(id email name created deactivated) },
+		});
+	}
+
+	$c->log->debug('user ' . $c->stash('user')->name . ' deactivating user ' . $user->name);
```
sungo: bump to warn please

lib/Conch/Controller/User.pm
karenetheridge: how about info here - as warn is more "something went differently than expected".

sungo: fine. Same in User.pm please

karenetheridge:
ok, lots more logging added.

lib/Conch/Controller/User.pm
```diff
@@ -219,6 +216,139 @@ sub change_password ($c) {
...
+	my $user_param = $c->stash('target_user');
+	my $user =
+		is_uuid($user_param) ? $c->db_user_accounts->find({ id => $user_param })
+	  : $user_param =~ /^email\=/ ? $c->db_user_accounts->find({ email => $' })
+	  : undef;
+
+	# user not found, but secure applications don't let on to the user
+	if (not $user) {
+		$c->log->warn("user lookup for $user_param failed");
+		return $c->status(202);
```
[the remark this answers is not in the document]
perigrin: However if the above were _not_ the case, I'd suggest throwing a 204 so that it's exactly the same response as the normal workflow. Otherwise they can sniff a 202 vs 204 and know which users were not found.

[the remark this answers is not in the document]
karenetheridge: that makes sense.  I will change those two bailouts back to:

```
return $c->status(404, { error => "user $user_param not found" }) if not $user;
```

karenetheridge:
one more force push (I only changed two 202 responses that are coming from authenticated-with-global-admin routes.)

Issue #141, opened 2018-02-07
User emails are case sensitive and shouldn't be


sungo:
We are currently dealing with email address in a case sensitive fashion. This allows us to create users with address like foo@joyent.com *and* foo@Joyent.com. There's already a production instance of this.  We need to make email addresses case insensitive.

sungo:
Fixed by #319

PR #319, opened 2018-08-02
fix: treat email addresses as if they were case insensitive for lookups

karenetheridge:
this closes issue #141.

This can also be achieved by changing user_account.email from type
'text' to type 'citext', and adding this to the schema (but it requires
superuser privileges to execute, so migrating might be awkward):

    CREATE EXTENSION IF NOT EXISTS citext;

### DIFF
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -279,7 +279,10 @@ sub create ($c) {
 
 	# we don't use lookup_by_* because they only search active users.
 	if (my $user = $c->db_user_accounts->search({
-			-or => { name => $name, email => $email }
+			-or => [
+				{ name => $name },
+				\[ 'lower(email) = lower(?)', $email ],
+			]
 		})->first)
 	{
 		return $c->status(409, {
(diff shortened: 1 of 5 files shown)


lib/Conch/Controller/User.pm
```diff
@@ -279,7 +279,10 @@ sub create ($c) {
 
 	# we don't use lookup_by_* because they only search active users.
 	if (my $user = $c->db_user_accounts->search({
-			-or => { name => $name, email => $email }
```
sungo: For existing users with multiple accounts based on case sensitivity, this will return multiple records.

karenetheridge:
> The migration also needs to introduce the new extension.

I'm not using the extension at all right now (instead, using `LOWER()` in sql), but if it's not an issue using that extension in the db then changing the schema is certainly easier.

I'll check the staging and prod dbs for existing dupes and ping bdha about what to do about them, and remove/fix them manually.

karenetheridge:
I added a db migration to remove the duplicate entries (after making a backup of the table), and some extra comments re what happens if `lookup_by_email` finds more than one result.

PR #323, opened 2018-08-03
Cleanup user references first

bdha:
Need to cleanup references or the migration fails to run.

```
conch=> delete from user_workspace_role where user_id in (  select distinct u1.id from user_account u1 inner join user_account u2 ON lower(u1.email) = lower(u2.email) and u1.created < u2.created );
DELETE 1
conch=> delete from user_account where id in ( select distinct u1.id from user_account u1 inner join user_account u2 ON lower(u1.email) = lower(u2.email) and u1.created < u2.created );
DELETE 1
```

### DIFF
--- a/cpanfile.snapshot
+++ b/cpanfile.snapshot
@@ -763,6 +763,21 @@ DISTRIBUTIONS
       curry 1.000000
       namespace::clean 0.23
       perl 5.008001
+  DBIx-Class-TimeStamp-0.14
+    pathname: R/RI/RIBASUSHI/DBIx-Class-TimeStamp-0.14.tar.gz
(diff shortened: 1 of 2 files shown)


karenetheridge:
I think there are other join tables that should also be checked at the same time. I'll look!

(BTW your PR brought unwanted entries back into cpanfile.snapshot.)

karenetheridge:
If the user-to-be-deleted has access to certain workspaces, should we grant that access to the user record we are leaving behind (as the new canonical user)? or just let them get invited to the workspace again if they need it?

karenetheridge:
see #325.

PR #328, opened 2018-08-06
First round of cleanups and simplifications on the Validation system

sungo:


lib/Conch/Controller/ValidationPlan.pm
```diff
@@ -49,8 +49,7 @@ sub create ($c) {
 
 	$c->log->debug("Created validation plan ".$validation_plan->id);
 
-	# FIXME why is this not using the same redirect methodology as the rest of the api?
-	$c->status( 201, $validation_plan );
+	$c->status(303 => "/validation_plan/".$validation_plan->id);
```
karenetheridge: 201 is more correct, I think, so I'd be happy to arrange a time when we consistently change to the correct way and change -cli and -ui at the same time to match.

lib/Conch/Validation.pm
```diff
@@ -14,17 +14,6 @@ Conch::Validation - base class for writing Conch Validations
 	has category    => 'CPU';
 	has description => q/Description of the validation/;
 
-	# Optional schema to validate $input_data before `validate` is run.
```
karenetheridge: This was a useful feature, I think?  Although perhaps not done in the best way.  I wonder if we could still just give the user access to the json validator in this class, and (optionally) allow them to specify their own schema name to do validations against, as that can get a lot of checks done faster/more concisely than expressing the same thing in perl, before moving on to more sophisticated checks that can't be expressed nicely in the schema.

(I'm not proposing making a change now, but instead considering it for future enhancements.)

sungo: Let's see how the rework comes out first. I suspect validations are going to look very different in the near future.

lib/Conch/Controller/ValidationPlan.pm
sungo: Agreed. But with one exception, the whole API uses 303 and I'm opting for consistency over correctness at the moment.

t/model/ValidationState.t
```diff
@@ -175,7 +168,7 @@ subtest "run validation plan" => sub {
 		0, 'Validation plan should have no validations' );
 	my $new_state =
 		Conch::Model::ValidationState->run_validation_plan( $device->id,
-		$validation_plan->id, {} );
+		$validation_plan, {} );
```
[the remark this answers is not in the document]
karenetheridge: this comment was dumb and should be ignored.

PR #329, opened 2018-08-06
When resetting a user's password, force them to change it

karenetheridge:
- the user can no longer use any endpoint without logging in with the
  new password which is insecure, since it was in an email in plaintext)

- the user still cannot use any endpoint before changing his password,
  and existing session tokens (from the insecure login) will only work for
  10 minutes to allow him to do this.

### DIFF
--- a/lib/Conch/Controller/Login.pm
+++ b/lib/Conch/Controller/Login.pm
@@ -313,30 +313,11 @@ with their new password.
 
 sub reset_password ($c) {
 	my $body = $c->req->json;
+
 	return $c->status( 400, { error => '"email" required' } )
 		unless $body->{email};
 
-	# check for the user and sent the email non-blocking to prevent timing attacks
-	Mojo::IOLoop->subprocess(
-		sub {
-			my $user = $c->db_user_accounts->lookup_by_email($body->{email});
-
-			if ($user) {
-				my $pw = $c->random_string();
-				$user->update({ password => $pw });
-
-				$c->log->info('sending password reset mail to user ' . $user->name);
-				Conch::Mail::password_reset_email(
-					{
-						email    => $user->email,
-						password => $pw,
-					}
-				);
-			}
-		},
-		sub { }
-	);
-	return $c->status(204);
+	return $c->status(301, "/user/email=$body->{email}/password");
 }
 
 =head2 refresh_token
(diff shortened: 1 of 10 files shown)


sungo:
The tests failure is caused by an issue spinning up the second Test::Conch instance in the integration tests. I'm looking into it.

lib/Conch/Controller/Login.pm
```diff
@@ -95,35 +100,43 @@ sub authenticate ($c) {
...
 		}
-		else {
-			$c->status( 401, { error => 'unauthorized' } );
-			return 0;
+
+		if ($user->force_password_change) {
+			$c->log->debug('basic auth failed: password correct, but force_password_change was set');
+			return $c->status(401, {
+				error => 'unauthorized',
+				message => 'reset password at /user/me/password within 10 minutes',
```
sungo: I'd like this converted to something other than a text string, particularly one in a 'message' element. Clients would have to parse this message to figure out if they need to be concerned.

lib/Conch/Controller/Login.pm
```diff
@@ -213,10 +248,32 @@ sub session_login ($c) {
...
+		$user->update({
+			last_login => \'NOW()',
+			password => $c->random_string,	# ensure password cannot be used again
+		});
+		# password must be reset within 10 minutes
+		$c->session(expires => time + 10 * 60);
+
+		# conch-ui will now redirect the user to the change password form;
+		# conch-shell will print a warning to the user to change his password now.
+		return $c->status(200, {
```
sungo: Should we be returning a 200 OK when it's really not ok because they need to change their password? paging @perigrin

lib/Conch/Controller/Login.pm
perigrin: Add a Location header.

lib/Conch/Controller/Login.pm
perigrin: 303 with a Location header is probably better.

karenetheridge: It should be 2xx, because we *are* logging them in.. they just need to do something afterwards to keep using the system.

karenetheridge: Does a redirect make sense, when we don't have a page to send them to that gives them a form to fill out and submit?  Or would the -ui intercept the redirect and send them to the equivalent form on its end?

sungo: But as an API consumer, the only way I can know that is by parsing a free form 'message' attribute on the response payload.

karenetheridge: yeah, a freeform message doesn't make sense, i agree.

PR #330, opened 2018-08-07
new test for multiple db connections in a test

karenetheridge:
This shows that we can create multiple Test::Conches that all point to
the same db, each maintaining their own connection to it.


karenetheridge:
tests pass; this should be good to merge.

PR #333, opened 2018-08-08
more misc background refactoring

karenetheridge:
debugging aids, and prereq cleanup.


Makefile
```diff
@@ -8,6 +8,9 @@ morbo: ## Run under morbo, listening on :5001
 
 build: local ## Install deps (TODO: and build docs)
 
+clean:
+	\rm -rf local log public/doc
```
perigrin: why not `git clean -dxf`?

karenetheridge: That would clean way way too much. e.g. right now in my repo, that would also take out:

  * all .swp files (of files I am currently editing)
  * persistent files in .gitignore that should always stay (.ackrc, .vimlocal, conch.conf)

PR #337, opened 2018-08-13
remove Conch::Model::Workspace(User)?

karenetheridge:
removes these modules, converting queries to DBIC:
* Conch::Model::Workspace
* Conch::Model::WorkspaceUser
* Conch::Class::Workspace
* Conch::Class::WorkspaceUser

..and makes some workspace routes more efficient.


lib/Conch/Controller/WorkspaceRoom.pm
```diff
@@ -55,15 +53,20 @@ sub replace_rooms ($c) {
 		});
 	}
 
-	if ( $workspace->name eq 'GLOBAL' ) {
+	my $uwr = $c->stash('user')->search_related('user_workspace_roles',
```
sungo: Unless I'm misreading, we do this query a lot. Should we abstract it?

lib/Conch/Controller/WorkspaceUser.pm
```diff
@@ -89,12 +98,14 @@ sub invite ($c) {
 		$c->log->warn("Email sent to ".$user->email." containing their PLAINTEXT password");
 	}
 
-	Conch::Model::Workspace->new->add_user_to_workspace(
-		$user->id,
-		$c->stash('current_workspace'),
-		$body->{role},
-	);
-	$c->log->info("Add user ".$user->id." to workspace ".$c->stash('current_workspace')->id);
+	# FIXME! do not downgrade a user's existing access to this workspace.
```
sungo: Do we have a way in the API to change a user's role other than `invite`? If not, `invite` needs to retain the ability to overwrite a user's permissions.

lib/Conch/Plugin/AuthHelpers.pm
```diff
@@ -30,17 +30,13 @@ workspace
 			my ( $c, $role_name ) = @_;
 			return 0 unless $c->stash('user_id');
 
-			my $ws = Conch::Model::Workspace->new->lookup_by_name('GLOBAL');
-			return 0 unless $ws;
+			# FIXME: currently does an exact match. should we return true
```
sungo: We need the ability to look up exact matches of permissions. If we want fuzzy matches, we can add a different helper for that.

sungo: (No change needed. Just a comment)

lib/Conch/Controller/WorkspaceUser.pm
sungo: (No change needed. Just a comment)

lib/Conch/Controller/WorkspaceRoom.pm
karenetheridge: I thought about putting a resultset for it into the stash, but decided to wait and see...

another possibility is writing a bunch of helper subs like 'current_user', 'current_workspace', 'current_device' etc that leveraged the \*_id path arguments. again I wanted to stare at the system a little bit from 1000 feet to see if the highways made better sense somewhere else.

lib/Conch/Plugin/AuthHelpers.pm
karenetheridge: I was writing helpers for that last week (e.g. `has_at_least_rw($workspace)`) but decided to hold off to see what we wanted to do with the changes with cascading perms.

lib/Conch/Controller/WorkspaceRoom.pm
sungo: If I'm counting right, in this PR, we use that same query six or seven times. I suspect our usage will only grow. So, one way or another, I'd like to abstract it.

PR #340, opened 2018-08-16
Remove Conch::Model::WorkspaceDevice

karenetheridge:
...and associated stored functions.


json-schema/response.yaml
```diff
@@ -18,10 +18,14 @@ definitions:
         type: string
   DetailedDevice:
     type: object
-    additionalProperties: false
+    additionalProperties:
```
sungo: additionalProperties is a boolean that tells the parser whether or not to allow properties that we did not describe. New properties need to go under the 'properties' element.

karenetheridge: additionalProperties is a boolean *or* an object. I think this is a more appropriate place for "stuff that may be there, but we don't care about it in any way whatsoever."

sungo: We have an immense amount of schema that does not use additionalProperties in this fashion. I'd prefer to maintain the current usage until we're ready to PR a global change.

karenetheridge: I checked again and I was using additionalProperties wrong (it *still can be an object*, but it specifies generic schemas for additional things, not listing them by name).
I have pushed an amendment.

PR #346, opened 2018-08-21
Changes required to support the SmartOS Reporter

bdha:


bdha:
In theory this is now functional, but tests are failing, so I suspect a monkey needs a banana.

Issue #351, opened 2018-08-24
Urgent fixes needed for conch-reporter-smartos support


bdha:
We need to get the new reporter deployed asap. A couple things I've run into so far:

```
# cat /tmp/conch-report.json | json conch | ./bin/conch -c /var/tmp/.conch.json api post /device/5T21TD2 -
{"error":true,"message":"HTTP Error: Status: 400 Bad Request\nBody: {\"error\":\"\\/temp\\/exhaust: Missing property.\\n\\/temp\\/inlet: Missing property.\"}\n"}
```
Inlet/exhaust requires a firmware upgrade, which seemingly a bunch of machines will be missing. `json-schema/input.yml` needs to be updated so these fields are not required.

```
# cat /tmp/conch-report.json | json conch | ./bin/conch -c /var/tmp/.conch.json api post /device/5T21TD2 -
{"error":true,"message":"HTTP Error: Status: 409 Conflict\nBody: {\"error\":\"Hardware product SKU 'NotProvided' does not exist\"}\n"}
```

We need to provide the `legacy_product_name` lookup function, if a device hw profile cannot be found using `product_name` or `sku`.

So far that's what I've run into.

There may be more.

sungo:
These changes have been addressed in all the active branches.

PR #352, opened 2018-08-24
Adds a fallback to legacy product name in DeviceReport#process and removes two json schema requirements for device reports

sungo:
Addresses joyent/conch#351 - This PR applies to release/v2.15 exclusively. Additional PRs for 2.16 and master will follow later as appropriate

### DIFF
--- a/lib/Conch/Controller/DeviceReport.pm
+++ b/lib/Conch/Controller/DeviceReport.pm
@@ -28,36 +28,48 @@ sub process ($c) {
 	my $device_report = $c->validate_input('DeviceReport');
 	return if not $device_report;
 	my $raw_report = $c->req->body;
+	my $schema = $c->schema;
 
-	my $maybe_hw;
+	my $hw;
+	if ($device_report->{device_type} && $device_report->{device_type} eq "switch") {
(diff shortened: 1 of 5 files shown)


lib/Conch/Controller/DeviceReport.pm
```diff
@@ -28,36 +28,48 @@ sub process ($c) {
...
+			sku => $device_report->{sku}
+		});
+
+		if(not $hw) {
+			$c->log->debug("Could not find hardware product by SKU, falling back to legacy_product_name");
+		    $hw = $schema->resultset('HardwareProduct')->find({
+				legacy_product_name => $device_report->{product_name}
+			});
+
+			$hw or return $c->render(status => 409, json => {
```
perigrin: This should be a 404 no?

sungo: I'd vote not because the user is doing a `POST /device/:id` and a 404 might be interpreted to mean that the device doesn't exist. This is validating data inside the report itself.

perigrin: Fair enough

PR #353, opened 2018-08-24
Cleans up DeviceReport#process, introducing a legacy product name lookup along the way.

sungo:
This addresses joyent/conch#351 for master, also bringing in some cleanup I've been working on elsewhere.

### DIFF
--- a/lib/Conch/Controller/DeviceReport.pm
+++ b/lib/Conch/Controller/DeviceReport.pm
@@ -19,54 +19,101 @@ with 'Conch::Role::MojoLog';
 
 =head2 process
 
-Processes the device report using the Legacy report code base
+Processes the device report, turning it into the various device_ tables as well
+as running validations
 
 =cut
 
 sub process ($c) {
-	my $device_report = $c->validate_input('DeviceReport');
-	return if not $device_report;
 	my $raw_report = $c->req->body;
+	my $schema = $c->schema;
 
-	my $maybe_hw;
-
-	if ( $device_report->{device_type}
-		&& $device_report->{device_type} eq "switch" )
-	{
-		$maybe_hw = Conch::Model::HardwareProduct->lookup_by_name(
-			$device_report->{product_name}
-		);
-		return $c->status(409, {
-			error => "Hardware product name '".$device_report->{product_name}."' does not exist"
-		}) unless ($maybe_hw);
-
-	} else {
-		$maybe_hw = Conch::Model::HardwareProduct->lookup_by_sku(
-			$device_report->{sku}
-		);
-
-		return $c->status(409, {
-			error => "Hardware product SKU '".$device_report->{sku}."' does not exist"
-		}) unless ($maybe_hw);
-
+	my $unserialized_report = $c->validate_input('DeviceReport');
+	if(not $unserialized_report) {
+		return; # JSON validator will handle the proper error response
 	}
 
-	# Use the old device report recording and device validation code for now.
-	# This will be removed when OPS-RFD 22 is implemented
-	$c->log->debug("Recording device report");
-	my ( $device, $report_id ) = $c->_record_device_report(
-		$device_report,
-		$raw_report
+	# Make sure the API and device report agree on who we're talking about
+	if ($c->param('id') ne $unserialized_report->{serial_number}) {
+		return $c->render(status => 422, json => {
+			error => "Serial number provided to the API does not match the report data."
+		});
+	}
+
+	my $hw;
+	# Make sure that the remote side is telling us about a hardware product we understand
+	if ($unserialized_report->{device_type} && $unserialized_report->{device_type} eq "switch") {
+		$hw = $schema->resultset('HardwareProduct')->find({
+			name => $unserialized_report->{product_name}
+		});
+	} else {
+		$hw = $schema->resultset('HardwareProduct')->find({
+			sku => $unserialized_report->{sku}
+		});
+
+		if(not $hw) {
+		  	$hw = $schema->resultset('HardwareProduct')->find({
+				legacy_product_name => $unserialized_report->{product_name}
+			});
+		}
+	}
+
+	if(not $hw) {
+		return $c->render(status => 409, json => {
+			error => "Could not locate hardware product"
+		});
+	}
+
+	if(not $hw->hardware_product_profile) {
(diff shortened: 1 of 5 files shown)


lib/Conch/Controller/DeviceReport.pm
```diff
@@ -19,54 +19,101 @@ with 'Conch::Role::MojoLog';
...
 sub process ($c) {
-	my $device_report = $c->validate_input('DeviceReport');
-	return if not $device_report;
 	my $raw_report = $c->req->body;
+	my $schema = $c->schema;
 
-	my $maybe_hw;
+	my $unserialized_report = $c->validate_input('DeviceReport');
+	if(not $unserialized_report) {
+		return; # JSON validator will throw the proper error response
```
karenetheridge: technically nothing is "thrown" - the response has just already been set in $c->tx->res.

lib/Conch/Controller/DeviceReport.pm
```diff
@@ -19,54 +19,101 @@ with 'Conch::Role::MojoLog';
...
+		});
 	}
 
-	# Use the old device report recording and device validation code for now.
-	# This will be removed when OPS-RFD 22 is implemented
-	$c->log->debug("Recording device report");
-	my ( $device, $report_id ) = $c->_record_device_report(
-		$device_report,
-		$raw_report
+	my $existing_device = $schema->resultset('Device')->find({
```
karenetheridge: This is already available in `$c->stash('current_device')` (and in the act of looking that up, it also verified that the user has permission to see this device, by virtue of their workspace(s) and/or previous relay reports -- although perhaps in this case that second part of the verification is not wanted?)

lib/Conch/Controller/DeviceReport.pm
```diff
@@ -228,7 +231,7 @@ sub _record_device_report {
 			my %inactive_serials = map { $_->serial_number => 1 } @device_disks;
 
 			foreach my $disk ( keys %{ $dr->{disks} } ) {
-				$log->debug("Device $device_id: Recording disk: $disk");
+				$log->debug("Device $device->id: Recording disk: $disk");
```
karenetheridge: This will not render correctly. use `$device_id`.

lib/Conch/Controller/DeviceReport.pm
```diff
@@ -267,11 +270,11 @@ sub _record_device_report {
 			}
 
 			$dr->{disks}
-				and $log->info("Recorded disk info for Device $device_id");
+				and $log->info("Recorded disk info for Device $device->id");
```
karenetheridge: This will not render correctly. use `$device_id`.

lib/Conch/Controller/DeviceReport.pm
```diff
@@ -190,18 +193,18 @@ sub _record_device_report {
...
+					cpu_num             => $dr->{processor}->{count},
+					cpu_type            => $dr->{processor}->{type},
+					nics_num            => $nics_num,
+					dimms_num           => $dr->{memory}->{count},
+					ram_total           => $dr->{memory}->{total},
 				}
 			);
 
-			$log->info("Created Device Spec for Device $device_id");
+			$log->info("Created Device Spec for Device ".$device->id);
```
karenetheridge: This will not render correctly. use `$device_id` (which needs to be initialized above, in case a new device was created).

lib/Conch/Controller/DeviceReport.pm
```diff
@@ -214,11 +217,11 @@ sub _record_device_report {
 			) if $dr->{temp};
 
 			$dr->{temp}
-				and $log->info("Recorded environment for Device $device_id");
+				and $log->info("Recorded environment for Device ".$device->id);
```
karenetheridge: This will not render correctly. use `$device_id`.

lib/Conch/Controller/DeviceReport.pm
```diff
@@ -282,7 +285,7 @@ sub _record_device_report {
 
 				my $mac = uc( $dr->{interfaces}->{$nic}->{mac} );
 
-				$log->debug("Device $device_id: Recording NIC: $mac");
+				$log->debug("Device $device->id: Recording NIC: $mac");
```
karenetheridge: This will not render correctly. use `$device_id`.

lib/Conch/Controller/DeviceReport.pm
sungo: No it's not. `POST /device/:id` is not part of that `under`.

sungo: The patch to `Route/Device.pm` included in this PR actually moves that route above the `under` declaration to make it more clear that it stands alone. 

Current behavior requires this route to get called even if the device doesn't exist.

lib/Conch/Controller/DeviceReport.pm
sungo: Addressed

lib/Conch/Controller/DeviceReport.pm
sungo: Addressed

lib/Conch/Controller/DeviceReport.pm
sungo: Addressed

lib/Conch/Controller/DeviceReport.pm
sungo: Addressed

lib/Conch/Controller/DeviceReport.pm
sungo: Addressed

lib/Conch/Controller/DeviceReport.pm
sungo: Addressed

lib/Conch/Controller/DeviceReport.pm
karenetheridge: k

karenetheridge: should we still be verifying that the device, if it already exists, is under a workspace the user has permission to access?

sungo: That's a question for a different PR. The code prior to this PR did not make that sort of permissions check so neither does the code in the PR.

PR #358, opened 2018-08-24
regenerate the validation docs and provide the ability to do so in our dependencies and makefile

sungo:


sungo:
Updated to Pod::Github 0.04

PR #363, opened 2018-08-27
remove the /feedback route that did nothing useful

sungo:


sungo:
If customers want to send us feedback, they know who we are and can just send us email directly.

PR #367, opened 2018-08-27
new GET /device?<thing>=<value> endpoint

karenetheridge:
where thing = mac, ipaddr

resolves GH#356.


karenetheridge:
note: I could just have this 303 to the /device/:device_id endpoint, if everyone is happy with that. currently it does the "redirect" internally all in the same request as that seemed more straightforward.

sungo:
I think the 200 is fine but I defer to @perigrin

perigrin:
200 is fine

PR #395, opened 2018-08-29
strip device models

karenetheridge:
I would suggest you just look at the overall diff for this one, unless you have questions about specific decisions I've made.


karenetheridge:
pushed two tiny amendments to ensure that `updated` is always set to `NOW()` on row change.

karenetheridge:
sorry, the last commit just got changed because I spotted a rebasing error in Conch::DB::ResultSet::DeviceSetting.

PR #402, opened 2018-08-30
After deleting a user, that email address cannot be reused in a new user.

karenetheridge:
(sungo wrote this)
I was testing workspace invites in the shell. I created a user via /workspace/:id/invite and then deleted them via `DELETE /usr/email=<REDACTED>`. When I attempted to invite that same email address again, the API threw a 500 with 

`DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::Pg::st execute failed: ERROR:  duplicate key value violates unique constraint \"user_account_email_key\"\nDETAIL:  Key (email)=(<REDACTED>) already exists. [for Statement \"INSERT INTO user_account ( email, name, password_hash) VALUES ( ?, ?, ? ) RETURNING id\" with ParamValues: 1='<REDACTED>', 2='<REDACTED>', 3='<REDACTED2>'] at /usr/home/sungo/src/conch/bin/../lib/Conch/Controller/WorkspaceUser.pm line 89`

My naive assumption is that we need to adjust the constraint on the email field to only apply to users where deactived=null

### DIFF
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -317,14 +317,8 @@ sub create ($c) {
 	# this would cause horrible clashes with our /user routes!
 	return $c->status(400, { error => 'user name "me" is prohibited', }) if $name eq 'me';
 
-	# we don't use lookup_by_* because they only search active users.
-	if (my $user = $c->db_user_accounts->search({
-			-or => [
-				{ name => $name },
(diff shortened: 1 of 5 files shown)


sungo:
So, I don't *think* this is a problem but I want to make sure I note the behavior some place. 

With my request, inviting a deactivated user reactivates the "old" account. That means that any other workspace assignments will spring back to life as well.

Issue #397, opened 2018-08-29
Remove `device_validate` table


bdha:
No longer used. Two steps to retire it:

- [x] Remove all code that references it (2.18)
- [x] Migration to drop the table (2.19)

karenetheridge:
and presumably device_validate_criteria too?

bdha:
Burn anything that moves or looks at ya funny.

karenetheridge:
well that's kinda everything...

karenetheridge:
phase 1: https://github.com/joyent/conch/pull/404

karenetheridge:
code for both phases is done. phase 1 is merged in #404; phase 2 PR to be submitted pending v2.19 opening up.

karenetheridge:
phase 2: #413

karenetheridge:
This is done in #404 and #413, complete in v2.19.

PR #404, opened 2018-08-31
burn device_validate, device_validate_criteria tables, phase 1

karenetheridge:
code removal only.

for #397


sungo:
We've not used those manta jobs in 10 months and they were only ever run that once, to build the data for last year's MBOs. If we decide we need them down the road, we can always pull them out of git history.

PR #406, opened 2018-08-31
add serial_number, asset_tag to datacenter_rack

karenetheridge:
..and add a mechanism for populating these fields via:
    
    POST /rack/:rack_id
    POST /workspace/:workspace_id/rack


karenetheridge:
looks like some of my tests got lost in the rebase. will fix next week.

karenetheridge:
ready for review+merge.

lib/Conch/Controller/Datacenter.pm
```diff
@@ -19,25 +17,20 @@ sub find_datacenter ($c) {
...
 
-	my $s;
-
-	if($c->stash('datacenter_id') =~ /^(.+?)\=(.+)$/) {
-		$c->status('501');
-		return undef;
-	} else {
-		$s = Conch::Model::Datacenter->from_id($c->stash('datacenter_id'));
+	if ($c->stash('datacenter_id') =~ /^(.+?)\=(.+)$/) {
+		return $c->status('501');
```
perigrin: This is pre-existing code so out of scope for this PR but ... I have no clue what 501 here _means_. Typically it's used for when an endpoint doesn't support the HTTP verb being used (e.g. read only should throw a 501 for PUT/POST) ... in this case we're sniffing the URL for the magic query params and throwing the 501 if we find them? (also capturing regex but we don't use the capture?)

sungo: Other similar endpoints supported lookup by non id fields like 'email=foo' and 'name=bar'. If the key wasn't supported, it threw a 501, sort of indicating that we don't support that key.

perigrin: Bleh, I ... bleh. Ok. I reserve the right to propose something better later though.

karenetheridge: yes, please do. it doesn't really make much sense the way it is now.

karenetheridge:
we need another review clicky for this to be merged, as another fix got merged into master in the meantime.

Issue #389, opened 2018-08-28
remove ability to login via user_account.name


sungo:

karenetheridge:
Email only?  Presently we accept all of: user id, name and email.

I can't imagine we get many attempts to log in via user id though :)

sungo:
automated tools should probably use the uuid. We should allow the user to change their email address and the uuid is the only immutable handle we have on them.

karenetheridge:
we don't allow changing the email address yet, but that's easy to do via `POST /user/me/email` whenever we want it (presumably not this week though)

sungo:
I expect we will/should so I'd like to preserve the ability to lookup by id as well as email

karenetheridge:
code is done; PR to be submitted pending other things being merged first.

karenetheridge:
This went into milestone 2.18 via #411 .

PR #411, opened 2018-09-04
remove ability to /login via user_account.name

karenetheridge:
for github #389

### DIFF
--- a/lib/Conch/Controller/Login.pm
+++ b/lib/Conch/Controller/Login.pm
@@ -211,24 +211,24 @@ Handles the act of logging in, given a user and password in the form. Returns a 
 =cut
 
 sub session_login ($c) {
-	my $body = $c->req->json;
-
-	return $c->status( 400, { error => '"user" and "password" required' } )
-		unless $body->{user} and $body->{password};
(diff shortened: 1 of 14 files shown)


lib/Conch/Controller/Login.pm
```diff
@@ -211,24 +211,24 @@ Handles the act of logging in, given a user and password in the form. Returns a
...
+		$c->log->debug('session login failed validation');
+		return $c->status(400);
+	}
 
 	# TODO: it would be nice to be sure of which type of data we were being passed here, so we
 	# don't have to look up by all columns.
-	my $user = $c->db_user_accounts->lookup_by_id($body->{user})
-		|| $c->db_user_accounts->lookup_by_name($body->{user})
-		|| $c->db_user_accounts->lookup_by_email($body->{user});
+	my $user = $c->db_user_accounts->lookup_by_id($input->{user})
```
perigrin: Do we ever supply the user's ID when logging someone in?

karenetheridge: No, but you can get the userid at `GET /user/me` and `GET /user/email=<foo>`.

PR #417, opened 2018-09-05
rename columns for rack unit size, position; add FK; adjust formatting of WorkspaceRack response

karenetheridge:
renames some db columns for clarity, and changes the formatting of the `GET /workspace/:workspace_id/rack/:rack_id` endpoint.

@perigrin this mostly impacts you, so I don't want to merge this until you've signed off on it.

### DIFF
--- a/json-schema/response.yaml
+++ b/json-schema/response.yaml
@@ -1337,7 +1337,7 @@ definitions:
       product_id:
         type: string
         format: uuid
-      ru_start:
+      ru_start: # TODO: this really should be rack_unit_start
         type: integer
       created:
         type: string
--- a/json-schema/response.yaml
+++ b/json-schema/response.yaml
@@ -621,41 +621,43 @@ definitions:
       datacenter:
         type: string
       slots:
-        type: object
-        additionalProperties: false
-        patternProperties:
-          ^[0-9]+$:
-            type: object
-            additionalProperties: false
-            required:
-              - alias
-              - id
-              - name
-              - size
-              - vendor
-              - occupant
-            properties:
-              alias:
-                type: string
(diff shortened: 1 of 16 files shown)


sungo:
The shell has support for this endpoint via `conch.GetWorkspaceRack`. It and the `Rack` struct will need to be adjusted.

karenetheridge:
If you want any other formatting changes to this endpoint (or any other endpoint for that matter), let's give'er.

sungo:
Not sure about renaming `ru_start` to `rack_unit_position`. That field is truly the starting point in the rack for the device. We list the device as being at RU 2, and it's a 4RU box. So it's position is 2 through 6.

@bdha what do you think?

bdha:
I would call it `rack_unit_start` if you want it to be more clear. Like @sungo says, it's really where the device starts in the rack.

karenetheridge:
ok, force-pushed with `s/rack_unit_position/rack_unit_start`.

json-schema/response.yaml
```diff
@@ -621,41 +621,43 @@ definitions:
...
-                description: Number of RUs for slot
-              vendor:
-                type: string
-                description: Hardware product vendor
-              occupant:
-                description: Device assigned to this slot or null
-                anyOf:
-                  - $ref: "#/definitions/Device"
-                  - type: null
+        type: array
```
perigrin: Please create a ticket in Conch-UI for this. It'll ultimately make life easier I think but right now it'll break everything all to hell.

karenetheridge: ok, as long as you're ok with resolving the breakage for 2.19!

PR #421, opened 2018-09-06
in Controller::DatacenterRack#update, if the user provides a "role" parameter, map it to the new db field name

sungo:


lib/Conch/Controller/DatacenterRack.pm
```diff
@@ -149,6 +149,10 @@ sub update ($c) {
 		}
 	}
 
+	if($input->{role}) {
+		$input->{datacenter_rack_role_id} = delete $input->{role};
+	}
+
```
karenetheridge: better (and one line):

`$input->{datacenter_rack_role_id} = delete $input->{role} if exists $input->{role};`

karenetheridge:
We should also have a test that covers this case, since clearly there isn't one presently.

I'll check the recent 2.17 patches to see that they have test coverage too.

Issue #400, opened 2018-08-29
Remove workspace validation-states API


bdha:
It is very expensive, and only the UI current uses it -- in a way that we want to remove.

See: https://github.com/joyent/conch-ui/issues/15

sungo:
This can't be processed until both the UI and the shell have removed their various functions and deployed to production.

karenetheridge:
code is done; PR to be submitted pending blockers being removed.

karenetheridge:
done for 2.19 in #425.

PR #425, opened 2018-09-11
remove endpoint: GET /workspace/:workspace_id/validation_state

karenetheridge:
for github #400


lib/Conch/Route/Workspace.pm
```diff
@@ -102,9 +101,6 @@ sub workspace_routes {
         $with_workspace->get('/user')->to('workspace_user#list');
         # POST /workspace/:workspace_id/user
         $with_workspace->post('/user')->to('workspace_user#invite');
-
-        # POST /workspace/:workspace_id/validation_state
-        $with_workspace->get('/validation_state')->to('workspace_validation#workspace_validation_states');
```
perigrin: it's a good thing we're removing this, the docs are wrong! (GET not POST)

karenetheridge: doh, commit message and PR label adjusted

PR #435, opened 2018-09-14
add unique index on device_settings.device_id+name

### LINKED ISSUE #433: device_settings queries are slow
0.5s in prod even for simple queries.  the `device_settings` table has nearly a million rows -- we need a unique index on the `name` field.

karenetheridge:
closes #433.


karenetheridge:
confirmed that all device_settings query constrain `deactivated is null`, so the index will get used.

PR #436, opened 2018-09-14
Revert "add unique index on device_settings.device_id+name"

sungo:
Reverts joyent/conch#435


karenetheridge:
will reset in master instead.

PR #450, opened 2018-09-18
remove all uses of IO::All

### LINKED ISSUE #445: remove IO::All from the codebase
see https://github.com/ingydotnet/io-all-pm/issues/70, https://github.com/ingydotnet/io-all-pm/issues/100.

karenetheridge:
closes #445.

### DIFF
--- a/lib/Test/ConchTmpDB.pm
+++ b/lib/Test/ConchTmpDB.pm
@@ -5,7 +5,7 @@ use warnings;
 use Test::PostgreSQL;
 use DBI;
 use Conch::DB;
-use IO::All;
+use Path::Tiny;
 
 use Exporter 'import';
 our @EXPORT_OK = qw( mk_tmp_db pg_dump );	 # TODO: do not export OO methods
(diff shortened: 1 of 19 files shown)


lib/Test/ConchTmpDB.pm
```diff
@@ -40,8 +40,8 @@ sub mk_tmp_db {
 		$dbh->do('CREATE EXTENSION IF NOT EXISTS "uuid-ossp";') or die;
 		$dbh->do('CREATE EXTENSION IF NOT EXISTS "pgcrypto";') or die;
 
-		for my $file ( io->dir("sql/migrations")->sort->glob("*.sql") ) {
-			$dbh->do( $file->all ) or die;
+		for my $file (sort glob('sql/migrations/*.sql')) {
```
perigrin: would this maybe be easier as `$dbh->do($_->slurp_utf8) or BAIL_OUT('Test SQL load failed') for sort path('sql/migrations)->children('*.sql')`?

t/model/DeviceLocation.t
```diff
@@ -2,12 +2,11 @@ use Mojo::Base -strict;
 use Test::More;
 use Test::ConchTmpDB qw(mk_tmp_db);
 
-use IO::All;
-
 use_ok("Conch::Model::DeviceLocation");
 
 use Data::UUID;
 use Conch::Pg;
+use Path::Tiny;
```
perigrin: You don't actually use `Path::Tiny` here.

karenetheridge: ah yes. I was, and then I removed the sql loading in this file because it's useless.

lib/Test/ConchTmpDB.pm
karenetheridge: I would meh, but I do usually like removing unnecessary additional scopes.

PR #451, opened 2018-09-18
convert ghch from LWP to HTTP::Tiny

### LINKED ISSUE #446: convert README extractor from LWP to HTTP::Tiny
..and remove extra deps that got added to cpanfile.snapshot.

see 9b4d0206.

karenetheridge:
...and repair cpanfile.snapshot
closes #446.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -16,7 +16,7 @@ requires 'Submodules';
 requires 'JSON::Validator';
 requires 'Pod::Github', '>= 0.04';
 requires 'Data::Validate::IP';      # for json schema validation of 'ipv4', 'ipv6' types
-requires 'LWP';
+requires 'HTTP::Tiny';
 
 # mojolicious and networking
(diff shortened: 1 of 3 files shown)


cpanfile
```diff
@@ -5,7 +5,7 @@ requires 'perl', '5.26.0';
 
 # basics
 requires 'Data::UUID';
-requires 'Data::Printer';
+requires 'Data::Printer', '0.99_019', dist => 'GARU/Data-Printer-0.99_019.tar.gz';
```
perigrin: Why a dep on an unreleased version? Are we using DDP for actual business logic?

karenetheridge: As described in the commit message, to get off of File::HomeDir and Mac::SystemDirectory.

We only use Data::Printer in development debugging.  It is `use`d in a few `lib` files, but not actually called. All references to it can likely be removed safely, but I suspect some contributors might still want it to be installed. (Personally I use Data::Dumper during development, so I'm ambivalent.)

perigrin: I'd personally remove it from the deps and then let people install it locally as they see fit.

PR #469, opened 2018-09-27
remove user creation from workspace invitation flow

### LINKED ISSUE #452: workspace invitation should require the user to already exist
Originally, the `POST /workspace/:id/user` endpoint was the only way to create a user. Now, we also have a separate user creation interface.  The workspace invitation endpoint should not create the user if it does not already exist, but instead return an error.

Needs coordination with the shell to modify existing workflows.

karenetheridge:
Email template adjusted accordingly.
This also gives us the bonus that newly created users will get a proper "name" entry.
Identifiers renamed to remove mention of "invitations".

Note that the "send_invite_mail" query parameter is now "send_mail".

closes #452.

### DIFF
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -336,7 +336,7 @@ definitions:
         anyOf:
           - type: 'null'
           - type: string
-  WorkspaceInvite:
+  WorkspaceAddUser:
     type: object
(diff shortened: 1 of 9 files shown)


karenetheridge:
rebased.

PR #477, opened 2018-10-01
Fix Session Expiration

perigrin:
JWT Tokens are set to expire in 1 or 30 day timeframe, but the implementation currently splits them and stores the signature in a cookie that expires in an hour. 

This PR fixes session expiration on the signature cookie to match the JWT Token.

It should be back-portable to v2.19.x


perigrin:
I thought this *was* rebased against master, I've updated my local master and rebased again.

PR #494, opened 2018-10-11
Location headers in redirects can be relative, to avoid forcing to the same proxy

karenetheridge:
and fix tests to check exact header content


sungo:
Please backport this into 2.19 and 2.20.

PR #500, opened 2018-10-12
do not require user_account.name to be unique

### LINKED ISSUE #496: A user's "name" should not be unique
The current database and code require both a user's email address and 'name' to be unique amongst active users.

The email address should be the sole non-UUID unique identifier for a user. The 'name' field should be used for pretty UI stuff alone since there are bound to be duplicates. We don't need to care what a user calls themselves, just that they have a unique email address where we can send password resets and what not.

Further, if login is accepting the 'name' field as a valid identifier, it should stop that.

karenetheridge:
no longer using name as an input identifier or authentication.
Note this will break any current users of basic auth whose 'name' has been changed to not equal 
their 'email', but there are few of these users, and we don't expect basic auth to be much in
use anymore.

closes #496.

### DIFF
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -411,8 +411,8 @@ definitions:
       - role
     properties:
       user:
-        description: user name or email
         type: string
+        format: email
       role:
         type: string
         enum:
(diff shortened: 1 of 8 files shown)


sungo:
Let's do some log analysis to make sure no one is using basic auth before killing it. We should at least provide a (short?) deprecation window

karenetheridge:
I've amended the comment to be more accurate. Basic auth should continue to work as long as the user's "name" isn't updated.

karenetheridge:
grepping of the last few days of logs on proxy0 shows no use of basic auth.

PR #505, opened 2018-10-16
more updates for DeviceReport json schema

karenetheridge:
..also fixes some errors made in #499 - some ints are being sent as strings.


sungo:
Are these new restrictions crucial to the API's existing usage of the device reports?

karenetheridge:
All of these changes reflect requirements the api has at some point in the flow when using this data, e.g. restrictions in the database schema where the data is stored. None of them are "this looks like what the data is now so let's just describe it".

PR #521, opened 2018-10-29
properly share db connections

karenetheridge:
A single DBIx::Class::Schema object performs its own connection management (e.g. reconnecting
as needed, managing savepoints and performing rollbacks), so the same object must be used
consistently for this all to work. Therefore we must cache the object in the application and
hand out the same object from the handler method, rather than creating new ones on every call
to the handler.


karenetheridge:
This should now be review- and merge-ready. It should go all the way back into v2.20 as transactions are currently compromised.

PR #522, opened 2018-10-31
Provide infrastructure to build and test with Docker, including scripts for buildbot

sungo:


sungo:
I think master/v2.22 is fine. If the 2.21 process takes a lot longer, we can always pull it over to do our first docker based release over there. But for now, I think it's fine to just build in docker from here on out.

PR #523, opened 2018-10-31
extend POST /hardware_product/(:id)? to create or update a hardware_product_profile

### LINKED ISSUE #511: Add endpoint which will allow hardware_product_profile insert/update from api
At the moment this is blocking me from adding any legacy device as quite a few servers and network switches/appliances will need their own profiles

karenetheridge:
closes #511.


sungo:
The docker test failure isn't spurious in this case. There's what appears to be a legit test failure in that environment. 

https://svc.conch.joyent.us/buildbotd/#/builders/4/builds/56/steps/2/logs/stdio

sungo:
Docker failure was caused by the Dockerfile pulling in postgresql 10. We're only certified against 9.6 so I modified the Dockerfiles to pull in 9.6 from the postgres repo.

bdelano:
Yay it works!  I was able to add a profile and a device, so im happy

PR #525, opened 2018-11-01
put the nytprof plugin behind a feature flag

sungo:


sungo:
Less breakage. It has its potential uses and it seemed ... nicer to just turn it off rather than purge it.

PR #529, opened 2018-11-02
add request id to response headers

### LINKED ISSUE #518: add request-id to the response headers
Since we are using bunyan logging and adding request id to the logs -- we should also send this information back in the response headers so when running verbose logs on the client you can trace the actual request in the api.

karenetheridge:
closes #518.


sungo:
Will wait for @jemershaw before merging, since this is his ask

PR #531, opened 2018-11-02
gracefully handle the lack of a git repository

karenetheridge:
On the other hand, maybe we want to make the lack of a properly-configured git repository fatal, so it's more obvious that something went wrong with the setup?

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -19,6 +19,7 @@ requires 'Data::Validate::IP';      # for json schema validation of 'ipv4', 'ipv
 requires 'HTTP::Tiny';
 requires 'Safe::Isa';
 requires 'Encode', '2.98';
+requires 'Capture::Tiny';
 
 # mojolicious and networking
 requires 'Mojolicious', '7.87'; # for Mojo::JSON's bootstrapping of Cpanel::JSON::XS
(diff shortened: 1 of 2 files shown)


sungo:
iirc, 'describe' doesn't work well / at all if you do a shallow clone

sungo:
Our current version string is semver. Do we want the unknown one to parse as semver too but some wacky number?

karenetheridge:
> Do we want the unknown one to parse as semver too

I lean towards 'no'.  I'm leaning more towards doing a `$c->log->fatal` there actually, since if the git repository was set up incorrectly, who knows what other parts of the deployment went wrong too?

sungo:
@karenetheridge are you still intending to make a lack of sensible git setup a fatal failure? I'm fine with that, for now. If we don't make it fatal, I'd like the version to parse as semver, but something ridiculous like 'v0.0.0-dirty'.

karenetheridge:
> are you still intending to make a lack of sensible git setup a fatal failure?

If it is our expectation that sensible git is part of the correct deployment process, then I'd rather we die, so it's clear early on what happened (no one looks at logs).  This might change in the future e.g. docker or whatever but it's easy to modify later.

PR #533, opened 2018-11-02
Always save device reports, flagging them as 'invalid' if necessary.

### LINKED ISSUE #377: Raise flag on unparseable reports
If a host is submitting reports we can't parse, we want to flip a bit somewhere and display that in the UI.

This helps with diagnosing service problems.

A device setting would work for this.

karenetheridge:
A new 'report_invalid' flag is returned with GET /device/:id to indicate that the last report
received for that report was invalid.

closes #377.


karenetheridge:
> If I understand the changes correctly, if I call GET /device/:id I can now get back an HTTP 200 with a DeviceReport, null, or an object like { "invalid": "yolostring" }. I understand the desire to provide access to invalid reports.

Would it be more sensible to mirror the previous functionality, returning the last *valid* report instead?

karenetheridge:
Alternatively, we could give the caller a choice with `?valid=1`, `?valid=0` to specify which type of report they want to see.  That would probably have to be `?invalid=1` to allow "no params" to keep doing what it did before (i.e. default to `?invalid=0`).

sungo:
Ok. So, I misread, so to clarify and correct myself (moar coffee). With this PR, we'd return a Device with a now-variable structure as LatestReport, content determined by the `report_invalid` flag. The shell can't easily use that field to determine how to parse that field. I'd need to specify that field as an empty interface (think 'yolo' as a data type) and then try to cast it again. It's messy and prone to unexpected behavior.

Thinking through things again, the PR also creates an extra problem for the device report field. We already struggle because that jsonb has multiple divergent report structures. This PR means we now have report-like structures and this new `invalid` structure in there. The field basically becomes a mongodb-esque unknowable blob to be parsed at runtime by upstream code. We've pushed for standardization of the reports so we can know what's in this field and this PR undoes those efforts.

I don't have a new design for you off the top of my head but, in my opinion, this PR creates more problems than it solves.

karenetheridge:
To avoid casting, we could return non-null results in one of two fields, depending on whether the most recent report was valid or invalid. I can't think of the right names for these though -- `latest_report` and `latest_report_invalid` are not strictly correct anymore if `latest_report` only contains a value if the latest report was valid.

karenetheridge:
I have updated this PR, updating the response payload of `GET /device/:id`:

*  'latest_report_is_invalid' flag is returned with GET /device/:id to indicate that the last report received for that report was invalid.
* 'latest_report' only contains a valid device report, as before (or null if the latest received report was not valid)
* 'latest_update' always contains the latest report received, whether it was valid or not (refer to 'latest_report_is_invalid' to know which)

sungo:
This just moves my concern to a different field. Yes, it prevents breakage in any existing code that uses `latest_report`. Now, `latest_update` has the parsing problem that I was discussing earlier so my concern still stands.

This also doesn't address my concerns about the contents of the device report jsonb field.

karenetheridge:
> Now, latest_update has the parsing problem that I was discussing earlier so my concern still stands.

Would you like this field (with a different name) to only be populated when it is an invalid report?

> This also doesn't address my concerns about the contents of the device report jsonb field.

Would you like failing reports to be stored in a separate column in the database?

karenetheridge:
I have pushed a new update that hopefully satisfies all concerns.

sungo:
Fix the test failure and I'll approve the PR.

PR #544, opened 2018-11-22
do not expose parent workspace ids to users who lack the auth to see them

### LINKED ISSUE #541: When requesting all workspaces, the API should not leak UUIDs that the user can't process
Currently, if I call `GET /workspace` to get a list of all my visible workspaces, I can get back workspaces that have parent IDs that I cannot access. A good example is a user that has access to 'us-east-1' but not 'GLOBAL'. The payload contains the relevant UUID but I get a 403 if I call `GET /workspace/:id` with that ID.

I'm not tied to any particular implementation but I do suggest that we not give the user IDs that they have no permission to access. Currently, we hand down a null if the workspace has no parent at all which may be appropriate for this use case. If the user has no access to the parent, maybe they shouldn't know there *is* a parent.

karenetheridge:
closes #541.

### DIFF
--- a/lib/Conch/Controller/Workspace.pm
+++ b/lib/Conch/Controller/Workspace.pm
@@ -103,6 +103,9 @@ sub get ($c) {
 		->with_role_via_data_for_user($c->stash('user_id'))
 		->single;
 
+	$workspace->parent_workspace_id(undef)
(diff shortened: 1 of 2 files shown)


sungo:
I suggested it in #541 so yeah, it works for me in the shell.

PR #549, opened 2018-11-27
refactor test scaffolding

karenetheridge:
This paves the way for many things to come, including finally removing the old-style database access and easier testing with sets of interdependent data.


lib/Test/Conch.pm
```diff
@@ -100,22 +110,59 @@ sub new {
...
 =head2 location_is
 
 Stolen from Test::Mojo's examples. I don't know why this isn't just part of the interface!
 
 =cut
 
-sub location_is {
-    my ($t, $value, $desc) = @_;
+sub location_is ($t, $value, $desc = 'location header') {
     $desc ||= "Location: $value";
```
perigrin: You have a default in the signature, this won't ever fire now will it?

karenetheridge: fixed; thanks.

Issue #461, opened 2018-09-25
Remove old device reports


sungo:
We have millions upon millions of old device reports. We need to figure out which ones to keep and which to burn and how to do so.

karenetheridge:
taking an initial stab, using pseudocode:

```
foreach device,
    consider all device reports, in reverse chronological order:
        if report has no associated validation_state record, see if there's another `device_report` row with identical json. if so, nuke it.
        if report passed, and the next report (the one we looked at in the previous loop iteration) also passed, then delete the `device_report` row,, its associated `validation_state`,  all `validation_result` rows etc.
        if report failed, see if its associated `validation_result` records all match the records for the report we looked at in the previous iteration. if so, delete the `device_report` row,, its associated `validation_state`,  all `validation_result` rows etc.

This is going to be *slow*.. we'll be stepping through all device_report records in the db and also doing some jsonb equality queries in the db.

sungo:
We probably can't use JSON equality as a decision point. Device reports contain temperature of the server and is often different between reports.

karenetheridge:
Do we care about preserving historical temperatures? If not, we can write a comparison routine that ignores that field.

sungo:
Paging @jemershaw

daleghent:
Ephemeral values for things such as system and disk temperatures (and perhaps in the future fan speeds or voltages... stuff like that) are ever-changing and are comparatively rarely a reason for us to flag a system for inspection in integration or pre-production. I think we can ignore those values. In production, a more appropriate venue for storing and alerting on these would be the regular production monitoring system (ie, Circonus and its agents) 

In this case, we can identify these elements of a report and ignore them for comparison reasons. They really are things that are useful in the moment they are generated and not much beyond that for our purposes.

sungo:
@daleghent we have #458 as well joyent/conch-relay#152 open for extracting that data from device reports and pivoting into a pluggable TSDB system.

daleghent:
+1 on offloading such data to a TSDB for cold storage. We only need the temps for validation purposes upon receipt of the report. Ignore them for the purposes of de-duping and go shrink that table :)

karenetheridge:
When comparing report json blobs for equivalence, what fields should I be ignoring? so far I just have (using perlish format for referencing fields):

* `{temp}`
* `{device}{*}{temp}`

Is there anything else?

karenetheridge:
> taking an initial stab, using pseudocode:

[...]

Issue #458, opened 2018-09-25
New endpoint for accepting time series data


sungo:
We need to be able to accept time series data from the relays and pivot it into some place useful. We initially need to support circonus, but should probably make it pluggable in case prometheus or something else comes into play.

The endpoint should accept an authenticated PUT or POST. No GET or DELETE is needed since we are just pushing data upstream.

@perigrin what are your thoughts on url structure?

karenetheridge:
(as discussed) one metric that would qualify here is temperature -- getting this out of device reports will help us de-dupe them.

karenetheridge:
Proposal:

    POST /device/:device_id/environment

json payload (inflated perl-style for readability):

    {
        temp => {
            cpu0 => 20,
            cpu1 => 20,
            inlet => 20,
            exhaust => 20,
        },
        voltage => {
            psu0 => 1.2,
            psu1 => 3.2,
        },
    }

Upon receipt of this data, we will:

* store it into `device_environment` (at least for now). all the above fields already have columns in this table, although we have never been processing/storing voltage data so far enven though some reports do contain this data.
* invoke an appropriate validation plan to check this data.  (what to do if validations fail is still undefined.)
* send it off to circonus (if the endpoint is ready), or log to a file for later import to circonus.

karenetheridge:
Now is a good time to provide feedback as to the format of the data coming into this endpoint, as well as the format of the data that will be logged as json (to be fed into circonus later on). It would be best if we mangle the data as little as possible between each layer.

sungo:
Superseded by #624

PR #550, opened 2018-11-27
Remove duplicate device reports; shunt environment data through a different endpoint

karenetheridge:
Not in a mergeable state yet. for discussion purposes.

See issues #458, #461.


karenetheridge:
closing, as we are going another way with device deletion (see #653). Environment data will be handled separately.

PR #556, opened 2018-11-29
check for occupancy conflicts when updating rack layouts

### LINKED ISSUE #427: signal when device-slot assignments are wrong
The `POST /workspace/:workspace_id/rack/:rack_id/layout` and `POST /device/:device_id/location` endpoints assign device(s) to specific rack locations. Currently there is no checking, either here or anywhere else, that this `device_location` mapping makes sense with the hardware that we have referenced in the `datacenter_rack_layout` table.  There should be a validation that makes this check.

Presently, the report in `t/integration/resource/passing-device-report.json` we use for testing even fails this check -- it is trying to store a "Test Storage" device in a slot intended for a "Test Compute".

karenetheridge:
...while taking into consideration that some hardware occupies more than one slot
    
closes #427.

### DIFF
--- a/lib/Conch/Command/check_layouts.pm
+++ b/lib/Conch/Command/check_layouts.pm
@@ -1,0 +1,68 @@
+package Conch::Command::check_layouts;
+
+use Mojo::Base 'Mojolicious::Command', -signatures;
+use Getopt::Long::Descriptive;
+
+=pod
+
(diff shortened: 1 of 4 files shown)


karenetheridge:
I found some existing conflicts in the staging db, which I gisted to ~Conch.

PR #573, opened 2018-12-06
add json schema entries for all data used by validations

### LINKED ISSUE #559: Check to make sure `dimms` and `cpus` are arrays and not objects.
@daleghent  pointed out on chat that we're occasionally getting single object values for dimms and cpus when they should be lists.

We currently have nothing that even checks that we're getting a dimms and cpus key in the Device Report. We probably should.

karenetheridge:
..and cpus, dimms, both of seem to be turning from arrays to hashes on some reporters
closes #559

### DIFF
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -87,6 +87,7 @@ definitions:
         type: string
 
   DeviceReport:
+    type: object
     required:
       - bios_version
       - memory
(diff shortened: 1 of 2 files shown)


sungo:
:+1:

karenetheridge:
Ideally this should be backported to 2.21 staging) and 2.20 (production) as well.

sungo:
I gave a +1 because the patch looks fine to me but the PR needs to be approved by the folks whose code will break probably. Specifically @daleghent and/or @jemershaw. This is all to catch a bug somewhere in the relay or reporter and the API will start rejecting reports that contain that bug.

Issue #580, opened 2018-12-12
DatacenterUpdate requires an 'id' in the payload


sungo:
Inconsistent to the other API endpoints, the DatacenterUpdate json schema, and thus `POST /dc/:id`, requires an ID in the payload.

PR #584, opened 2018-12-13
better datacenter endpoints

karenetheridge:
Adds numerous checks before mutating datacenter_* structures; fixes holes in interfaces.


sungo:
This appears to close #580

karenetheridge:
Individual commits note which issues they are closing. they include 580, 520, 567.

Issue #587, opened 2018-12-13
Want validation for hardware RAID LUNs


daleghent:
We need to be able to validate the number of RAID LUNs present on a system, such as those made with Avago MegaRAID cards.

Issue #586, opened 2018-12-13
Want validation for NVMe drives


daleghent:
Similar to #564, we need to be able to validate NVMe drives present in a system. These are being reported with a `drive_type` of `NVME_SSD`

Issue #564, opened 2018-12-05
Want validations for SATA HDDs


daleghent:
While the legacy `hardware_produce_profile` table allows for us to enumerate quantities of SATA drives that are in servers, no validation exists to count the number of `SATA_HDD` `drive_type`s being reported.

PR #590, opened 2018-12-14
Add missing drive type validators

daleghent:
Addresses:
#564 Want validations for SATA HDDs
#586 Want validation for NVMe drives
#587 Want validation for hardware RAID LUNs

* Migrates old SSD type to SATA SSD (`ssd_*` -> `sata_ssd_*`)
* Migrates old SATA type to SATA HDD (`sata_*` -> `sata_hdd_*`)
* Migrates old SAS type to SAS HDD (`sas_*` -> `sas_hdd_*`)
* Adds new type: SAS SSD (`sas_ssd_*`)
* Adds new type: NVMe SSD (`nvme_ssd_*`)
* Adds new type: RAID LUN (count only, `raid_lun_num`)
* Add test for RAID LUNs and SMART data
* Spelling and capitalization corrections


lib/Conch/DB/Result/HardwareProductProfile.pm
```diff
@@ -94,51 +94,86 @@ __PACKAGE__->table("hardware_product_profile");
   data_type: 'integer'
   is_nullable: 0
 
-=head2 sata_num
+=head2 sata_hdd_num
```
karenetheridge: The checksums haven't been updated in this file -- did you run 'make dbic', which is the only thing that should be modifying the upper portion of the file?

sql/migrations/0072-add-nvme_ssd_num.sql
```diff
@@ -0,0 +1,8 @@
+SELECT run_migration(72, $$
```
karenetheridge: All these migration files should be squashed into one.

sql/schema.sql
```diff
@@ -466,15 +466,22 @@ CREATE TABLE public.hardware_product_profile (
...
+    sas_hdd_num integer,
+    sas_hdd_size integer,
+    sas_hdd_slots text,
+    sata_ssd_num integer,
+    sata_ssd_size integer,
+    sata_ssd_slots text,
+    sas_ssd_num integer,
+    sas_ssd_size integer,
+    sas_ssd_slots text,
+    nvme_ssd_num integer,
```
karenetheridge: 'nvme_ssd_size', 'nvme_ssd_slots', 'sas_ssd_size' and 'sas_ssd_slots' are not used anywhere. Do we need to add them?

daleghent: Before the current validation system, `*_size` and `*_slots` columns were actually validated on. When Lane switched us to the current validation system early last year, he did so with the intention of re-implementing validations for those types of things, but left before he was able to get around to doing so. It's my intention to (eventually) re-implement validations for those things.

lib/Conch/Validation/SasSsdNum.pm
```diff
@@ -19,24 +19,15 @@ sub validate {
...
 
-	my $ssd_want = $hw_profile->ssd_num || 0;
-
-	# Joyent-Compute-Platform-3302 special case.  HCs can have 8 or 16 SSD and
-	# there's no other identifier. Here, we want to avoid missing
-	# failed/missing disks, so we jump through a couple extra hoops.
-	if ( $self->hardware_product_name eq "Joyent-Compute-Platform-3302" ) {
-		if ( $ssd_num <= 8 ) { $ssd_want = 8; }
-		if ( $ssd_num > 8 )  { $ssd_want = 16; }
-	}
```
perigrin: Won't this kill our ability to revalidate these devices? Does that matter?

lib/Conch/DB/Result/HardwareProductProfile.pm
daleghent:
```
daleg@debian9-vm1:~/conch$ git grep "make dbic"
daleg@debian9-vm1:~/conch$
```
ENODOCS on this, so no. I wasn't aware of this.

daleghent:
@karenetheridge I wasn't aware that we have commit message content standards. The merge message will reference the pull request which has all the detail. Is this not sufficient?

I think some history is in order:

As originally implemented in Conch, disk types were unfortunately over-generalized. This change, as drastic as it may appear to be, addresses this over-generalization in its entirety. Right now we have to lie in reports about disk types. As an example, we currently can't differentiate between NVMe, SAS, or SATA SSDs and we now have systems which possess a mix of both. Without the granularity that this change enables, we must generate reports and lie about disk types ([we report NVMe SSDs as SAS SSDs](https://github.com/joyent/conch-livesys/blob/master/Conch-Reporter/lib/Conch/Reporter/Collect/Disk.pm#L192), which obviously isn't right, but it's necessary to get _anything_ accomplished) and infer later what they mean at the risk of lost accuracy and missing any problems. It also means we are prevented from asking "Does this server have X number of the right kind of drives?" and are instead reduced to asking a decidedly less-detailed "Does this server have X drives?"

The renaming of the existing columns is so that clarity in their purpose and representation is assured and that they are consistent in name with the drive types this patch introduces. `ssd_num` alone is too ambiguous in the face of there being 3 distinct types of SSDs (SATA, SAS, and NVMe), and `sas_num` (and friends) and `sas_num` (and friends) are equally ambiguous on their own. I am willing to yield on renaming the existing columns if doing so poses some undue burden elsewhere, but doing that will retain some of the stench of technical debt that I'd much prefer to fully get rid of.

lib/Conch/Validation/SasSsdNum.pm
daleghent: This test has been moved to where[ it should be](https://github.com/joyent/conch/pull/590/files#diff-be4ef5f6bab66514b39768047effbd31R25) and the testing of reports from JCP-3302s should not be impacted.

sungo:
@daleghent will these renames render unusable the existing device reports in the database? For instance, after your changes, if I post any of our most recent device reports into the API, will they continue to validate as expected?

daleghent:
I'd also like to add that I'm open to suggestions and guidance in making this easier to be integrated. I'm not as aware of all the nooks and crannies as you all are, so while `make test` appears to work and basic functionality has the appearance of remaining intact, It's likely missing a few things that I have no awareness of.

karenetheridge:
> @karenetheridge is gonna go through and specifically see if she sees anything that will keep this from being back-ported to 2.21 (current production) 

Leaving aside the issue of renaming existing db fields, there will be *many* conflicts (there was a rewrite of all the hardware_product endpoints in v2.22) but it is possible, given sufficient bribes.

sql/schema.sql
lseppala: @daleghent sorry my friend! I remember trying to make the conversion to the new system exactly 1:1 with all the old validations. My apologies if I missed one and left more work for you to do.

Glad to see you extending the validation system in interesting ways!

sungo:
After talking this through in MM, we're going to target this PR for 2.23. We're going to tough it out for the immediate build at hand but want this code in production for an upcoming build in early Feburary. If 2.23 is not in production then, we can talk about backporting the PR to 2.22 which is more a discussion about risk than it is about code.

So, in that light, we don't need to worry about shell breakage. Just make a note of it and file a ticket for me in that repo. @karenetheridge breaks more than this for me in PRs all the time :)

As for technical debt, in a PR for 2.23, I don't see any problems tackling that now. I'm tempted to say let it ride until the planned major changes to the validation system but it doesn't hurt to tackle it now. 

As for my concern about breaking the validation of old device reports, I withdraw that concern. This will break our ability to revalidate previous reports. It's important to know that fact and that this PR is one of those lines.  However, revalidating old device reports is not something we do and not something we have code for. Our various test corpus will need updated with the new format, of course.

With all that said, let's redirect feedback and conversation to what is necessary to get the PR into a condition everyone is happy merging for 2.23/master.

PR #602, opened 2019-01-09
port scripts from conch-json-schema repo

perigrin:
Migrate the `validate.pl` and `extract-schema.pl` scripts from
the conch-json-schema repo as we've deprecated and will soon
remove that.


misc/extract-schema.pl
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env perl
+use 5.26.1;
+use feature 'signatures';
```
karenetheridge: `experimental` is in core, so you can just `use experimental 'signatures';`

misc/extract-schema.pl
```diff
@@ -0,0 +1,42 @@
...
+no warnings 'experimental::signatures';
+
+use Getopt::Long;
+use YAML::XS qw(LoadFile);
+use Mojo::JSON qw(encode_json);
+use Mojo::File qw(path);
+use Data::Visitor::Tiny qw(visit);
```
karenetheridge: This will need to be added to cpanfile*.

misc/extract-schema.pl
```diff
@@ -0,0 +1,42 @@
...
+
+my $schema_file = "$FindBin::Bin/../json-schema/input.yaml";
```
karenetheridge: DirSelf is available in cpanfile already and provides a less wtf way of getting this information.

perigrin: I've changed it but I'm not 100% sold that it's _really_ that much less of a wtf.

karenetheridge:
+1 from me. I'm happy for a squashed version of this to be merged.

PR #603, opened 2019-01-09
cpu information is not necessarily unique

karenetheridge:
..at least, not until we have more fields in the schema, that we know provide uniqueness.

### DIFF
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -101,7 +101,6 @@ definitions:
         type: string
       cpus:
         type: array
-        uniqueItems: true
         items:
           type: object
       dimms:


sungo:
@daleghent your approval of this PR is the important one.

For me, +1

karenetheridge:
Alternatively, I could add this:

```
        cpus: 
          type: array
         items: 
            type: object
          type: array
           items:
             type: object
+          required:
+            - processor
+          properties:
+            processor:
+              type: object
+              required:
+                - count
+              properties:
+                count:
+                  description: this should match hardware_product_profile.cpu_num
+                  type:
+                    integer
```

PR #611, opened 2019-01-11
moar tests

karenetheridge:
- refactored some common bits relating to logging in into a new Test::Conch method
- a few more tests or improvements to existing ones
- fixed broken bits in the sample device report, which start to cause a problem later on when we actually start caring harder about validations


sungo:
Not necessarily something for this PR but are the ping tests necessary? If the test API isn't working, we'll know pretty much on the next line of the test file.

PR #612, opened 2019-01-11
Serve Schemas from the API

perigrin:
Our API messages are all based around JSON Schemas. This exposes those schemas to clients and other consumers so they can validate the messaging themselves. 

Request schemas are exposed via:

    http://conch.joyent.us/schema/request/:schema_name

Response schemas are exposed via 

    http://conch.joyent.us/schema/response/:schema_name

There is currently no way to list the possible values of `:schema_name`, that's reserved for future work.

Also the `misc/validate` script has been updated to pull from the production URLs by default.

### DIFF
--- a/lib/Conch/Controller/Schema.pm
+++ b/lib/Conch/Controller/Schema.pm
@@ -1,0 +1,72 @@
+package Conch::Controller::Schema;
+
+use Mojo::Base 'Mojolicious::Controller', -signatures;
+
+use Data::Visitor::Tiny qw(visit);
+use Conch::Plugin::JsonValidator;
+use Mojo::Util qw(camelize);
+
+=pod
+
+=head1 NAME
+
+Conch::Controller::Schema
+
+=head1 METHODS
+
+=head2 get
+
+Get the json-schema in JSON format.
+
+=cut
+
+sub get ($c) {
+	my $type = $c->stash('request_or_response');
+	my $name = camelize $c->stash('name');
+
+	my $validator = JSON::Validator->new();
+
+	if ( lc($type) eq 'response' ) {
+		$validator->schema(Conch::Plugin::JsonValidator::OUTPUT_SCHEMA_FILE);
+
+	}
+	elsif ( lc($type) eq 'request' ) {
+		$validator->schema(Conch::Plugin::JsonValidator::INPUT_SCHEMA_FILE);
+	}
+
+	my $schema = $validator->get("/definitions/$name");
+
+	my sub inline_ref ( $ref, $schema ) {
+		my ($other) = $ref =~ m|#?/definitions/(\w+)$|;
+		$schema->{definitions}{$other} = $validator->get($ref);
+	}
+
+	visit $schema => sub ( $key, $ref, @ ) {
+		inline_ref( $_ => $schema ) if $key eq '$ref';
+		if ( !defined $_ && $key eq "type" ) {
+			$$ref = "null";
+		}
+	};
+	$schema->{title} //= $name;
+	$schema->{'$schema'} = 'http://json-schema.org/draft-07/schema#';
+	$schema->{'$id'}     = "urn:$name.schema.json";
+
+	return $c->status( 200, $schema );
+}
+
+1;
+__END__
+
+=pod
+
+=head1 LICENSING
+
+Copyright Joyent, Inc.
+
+This Source Code Form is subject to the terms of the Mozilla Public License,
+v.2.0. If a copy of the MPL was not distributed with this file, You can obtain
+one at http://mozilla.org/MPL/2.0/.
+
+=cut
+
+# vim: set ts=4 sts=4 sw=4 et :
--- a/lib/Conch/Route.pm
+++ b/lib/Conch/Route.pm
@@ -71,6 +71,10 @@ sub all_routes (
     $root->post('/logout')->to('login#session_logout');
     $root->post('/reset_password')->to('login#reset_password');
 
+    # GET /schema/:input_or_response/:schema_name
+    $root->get( '/schema/:request_or_response/:name' =>
+				[ request_or_response => [qw(request response)] ] )->to('schema#get');
+
     # GET /workspace/:workspace/device-totals
     $root->get('/workspace/:workspace/device-totals')->to('workspace_device#device_totals');
 
--- a/misc/validate
+++ b/misc/validate
@@ -2,30 +2,39 @@
 use 5.12.1;
 use warnings;
 
-use Dir::Self;
 use Getopt::Long;
 use JSON::Validator;
 use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
 use Pod::Usage;
 
-my $schema_file = __DIR__ . '/../json-schema/input.yaml';
-my $name        = 'DeviceReport';
+my $schema_url = 'https://conch.joyent.us/schema/request';
+my $name       = 'DeviceReport';
 
 GetOptions(
-	'file|f:s'   => \$schema_file,
-	'schema|s:s' => \$name,
-	'report|r:s' => \my $report,
-	'help|h'     => \my $help,
+	'base_url|u:s'    => \$schema_url,
+	'schema_file|f:s' => \my $schema_file,
+	'schema|s:s'      => \$name,
+	'report|r:s'      => \my $report,
+	'help|h'          => \my $help,
 );
 
 pod2usage(1) if $help;
 
 my $validator = JSON::Validator->new;
-$validator->schema($schema_file);
+
+if ($schema_file){
+    my $schema = path($schema_file)->to_abs;
+    $validator->load_and_validate_schema($schema);
+} else {
+    $validator->load_and_validate_schema("$schema_url/$name");
+}
+
 $validator->coerce(1);
 
-$report = defined $report ? path($report)->slurp : do { local $/; <> };
+$report = decode_json(
+	defined $report ? path($report)->slurp : do { local $/; <> }
+);
 
 my $schema = $validator->get($name);
 if ( my @errors = $validator->validate( $report, $schema ) ) {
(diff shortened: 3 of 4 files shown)


lib/Conch/Controller/Schema.pm
```diff
@@ -0,0 +1,60 @@
+package Conch::Controller::Schema;
+
+use Mojo::Base 'Mojolicious::Controller', -signatures;
+
+use Role::Tiny::With;
+with 'Conch::Role::MojoLog';
+
+use Mojo::Exception;
```
karenetheridge: this isn't needed?

lib/Conch/Controller/Schema.pm
```diff
@@ -0,0 +1,60 @@
...
+
+=head2 get
+
+Get the json-schema in JSON format.
+
+=cut
+
+sub get ($c) {
+	my $type = $c->param('input_or_response');
+	my $name = $c->param('name');
```
karenetheridge: Values pulled out of the path itself are better taken as `$c->stash($name)`, with param reserved for form parameters.

lib/Conch/Route.pm
```diff
@@ -71,6 +71,9 @@ sub all_routes (
     $root->post('/logout')->to('login#session_logout');
     $root->post('/reset_password')->to('login#reset_password');
 
+    # GET /schema/:input_or_response/:schema_name
+    $root->get('/schema/:input_or_response/:name')->to('schema#get');
```
karenetheridge: You may want to put the `(request|response)` restriction right here so the controller doesn't need to check it:  https://metacpan.org/pod/Mojolicious::Guides::Routing#Restrictive-placeholders

t/schema.t
```diff
@@ -0,0 +1,39 @@
+use strict;
+use warnings FATAL => 'utf8';
+use utf8;
+use open ':std', ':encoding(UTF-8)'; # force stdin, stdout, stderr into utf8
+
+use Data::UUID;
+use Path::Tiny;
+use Test::Conch;
+use Test::Deep;
+use Test::Memory::Cycle;
```
karenetheridge: I don't think you want this here?

t/schema.t
```diff
@@ -0,0 +1,39 @@
+use strict;
+use warnings FATAL => 'utf8';
+use utf8;
```
karenetheridge: I don't see any non-ascii content here?

t/schema.t
```diff
@@ -0,0 +1,39 @@
...
+use Test::Warnings ':all';
+use JSON::Validator;
+
+
+my $json_schema = JSON::Validator->new->schema('http://json-schema.org/draft-04/schema#')->schema->data;
+
+my $t = Test::Conch->new;
+$t->load_fixture('legacy_datacenter');
+
+$t->get_ok('/ping')->status_is(200)
```
karenetheridge: This test isn't needed here.

t/schema.t
```diff
@@ -0,0 +1,39 @@
...
+use Test::Memory::Cycle;
+use Test::More;
+use Test::Warnings ':all';
+use JSON::Validator;
+
+
+my $json_schema = JSON::Validator->new->schema('http://json-schema.org/draft-04/schema#')->schema->data;
+
+my $t = Test::Conch->new;
+$t->load_fixture('legacy_datacenter');
```
karenetheridge: You shouldn't need any fixtures for this test.

t/schema.t
```diff
@@ -0,0 +1,39 @@
...
+
+$t->get_ok('/ping')->status_is(200)
+	->json_is('/status' => 'ok')
+	->header_isnt('Request-Id' => undef)
+	->header_isnt('X-Request-ID' => undef);
+
+$t->get_ok('/schema/response/Login')
+       ->status_is(200)
+       ->json_schema_is($json_schema)
+       ->json_is('/type' => 'object')
```
karenetheridge: You may find `json_cmp_deeply` more convenient to use, to test more of the data structure all in one go. I found json_is to be really fiddly in having to specify just one bit at a time, leading to lots of little tests.

t/schema.t
```diff
@@ -0,0 +1,39 @@
...
+       ->json_is('/properties/jwt_token/type' => 'string');
+
+$t->get_ok('/schema/request/Login')
+        ->status_is(200)
+        ->json_schema_is($json_schema)
+       ->json_is('/type' => 'object')
```
karenetheridge: augh inconsistent whitespace :)

t/schema.t
```diff
@@ -0,0 +1,39 @@
...
+use utf8;
+use open ':std', ':encoding(UTF-8)'; # force stdin, stdout, stderr into utf8
+
+use Data::UUID;
+use Path::Tiny;
+use Test::Conch;
+use Test::Deep;
+use Test::Memory::Cycle;
+use Test::More;
+use Test::Warnings ':all';
```
karenetheridge: No import is needed for this file, as this is only being loaded for the 'had no warnings' hook in `done_testing`.

t/schema.t
```diff
@@ -0,0 +1,39 @@
+use strict;
+use warnings FATAL => 'utf8';
+use utf8;
+use open ':std', ':encoding(UTF-8)'; # force stdin, stdout, stderr into utf8
+
+use Data::UUID;
+use Path::Tiny;
```
karenetheridge: These two imports aren't needed either.

misc/validate
```diff
@@ -2,30 +2,37 @@
...
+    my $schema = path($schema_file)->to_abs;
+    $validator->load_and_validate_schema($schema);
+} else {
+    $validator->load_and_validate_schema("$schema_url/$name");
+}
+
 $validator->coerce(1);
 
-$report = defined $report ? path($report)->slurp : do { local $/; <> };
+$report = decode_json defined $report ? path($report)->slurp : do { local $/; <> };
```
karenetheridge: Can the read-from-stdin option be mentioned in usage?

lib/Conch/Controller/Schema.pm
```diff
@@ -46,6 +46,22 @@ sub get ($c) {
 	}
 
 	my $schema = $validator->get("/definitions/$name");
+
+    my sub inline_ref ( $ref, $schema ) {
```
karenetheridge: If I understand correctly, this is inlining the underlying definitions whenever there is a `$ref`? Could this be shown in a test?  I guess something like `$device_report_response->json_is('/system_uuid' => { type => 'string', pattern => '...' })`

misc/validate
perigrin: I'm not exactly sure how to document that in the usage. Got an example? It is explicitly mentioned that the report defaults to STDIN in the docs for `-r` which come up in the help.

lib/Conch/Controller/Schema.pm
perigrin: I'll make it explicit. It's implicit in the test for a device report but yeah I'll make an explicit test for it.

karenetheridge: Is it? I don't see where that's tested.

lib/Conch/Controller/Schema.pm
```diff
@@ -0,0 +1,60 @@
...
+
+use Mojo::Exception;
+use Data::Visitor::Tiny qw(visit);
+use Conch::Plugin::JsonValidator;
+
+=pod
+
+=head1 NAME
+
+Conch::Controller::User
```
karenetheridge: wrong package name

lib/Conch/Controller/Schema.pm
```diff
@@ -23,9 +23,14 @@ Get the json-schema in JSON format.
 
 =cut
 
+my sub camelize ($str) {
```
karenetheridge: BTW there is Mojo::Util::camelize

misc/validate
karenetheridge: ah, ok, I missed that.

lib/Conch/Controller/Schema.pm
perigrin: The test for the device report failed before I copied that code over from `misc/extract-schema` :)

I've added an explicit test now in the earlier test for the Login message to see if the `$ref` we have there is inlined.

lib/Conch/Controller/Schema.pm
perigrin: TIL.

lib/Conch/Controller/Schema.pm
perigrin: fixed in 584081e

t/schema.t
perigrin: Should be fixed in  bdf6f8d

t/schema.t
perigrin: Should be fixed in bdf6f8d

t/schema.t
perigrin: I re-ran perltidy, lemme know how you want it fixed.

t/schema.t
perigrin: I'm gonna leave these as-is since they're simple. I have used it for an expanded example.

t/schema.t
perigrin: Removed.

t/schema.t
perigrin: Removed in bdf6f8d

t/schema.t
perigrin: Removed in bdf6f8d

lib/Conch/Route.pm
perigrin: Fixed in fdeef31

lib/Conch/Controller/Schema.pm
perigrin: Fixed in fdeef31

lib/Conch/Controller/Schema.pm
perigrin: Removed.

t/schema.t
perigrin: removed.

misc/validate
karenetheridge: need moar parentheses? internal keywords have very high operator precedence and everything else has very low, so I *think* what you want is `decode_json(defined $report ? path($report)->slurp : do { local $/; <> });` ?

lib/Conch/Controller/Schema.pm
karenetheridge: This section (filling in the $refs) doesn't really go with the commit as it is currently labelled, but ok...

perigrin: Split this out into it's own commit.

misc/validate
perigrin: added moar parens.

PR #634, opened 2019-02-01
Fix DIMMCount and CPUCount

perigrin:
The Device Report sends up a `dimms` and a `cpus` properties that list the memory DIMMs and the CPUs respectively. These were intended to replace the older `memory` and `processor` properties. 

This PR updates the `DIMMCount` and `CPUCount` validations to use the newer properties instead of the older one. This entirely deprecates `processors` and leaves the only the `total` property being used from the object in the `memory` property (which currently has no analog in DIMMs).

### DIFF
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -90,8 +90,6 @@ definitions:
     type: object
     required:
       - bios_version
-      - memory
-      - processor
       - product_name
       - serial_number
       - state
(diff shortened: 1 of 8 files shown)


daleghent:
`{memory: {total: <int in GB>}` can be replaced by summing the value of `memory-size: <int in GB>` that is present in each `dimms` member. If a particular DIMM slot is unstuffed, there will be no such member (a recent change; empty members are now elided from the report). So basically, if `memory-size` exists, add its value to a running summation and when you've added them all up, you should have a number that is equal to what `{memory: {total: $value}` gives us now.

PR #641, opened 2019-02-05
Cut validations over to DBIC

karenetheridge:


lib/Conch/Validation.pm
```diff
@@ -51,308 +56,331 @@ functions tests that Validations define the required attributes and methods,
...
-=head1 METHODS
-
 =cut
 
 package Conch::Validation;
 
-use Mojo::Base -base, -signatures;
-use Mojo::Exception;
+use Moo;
+no Moo::sification;
```
perigrin: This isn't a problem with our codebase currently, but I had to do a bunch of jumping around in `Sqitch` to allow someone to use it in the same process space as a Moose class. We had to remove this line from that and replace it with a test (https://github.com/sqitchers/sqitch/blob/master/t/mooseless.t) so that it became _possible_.

Which is the longwinded way of me saying, I'd prefer to _test_ for accidental Moosification in our codebase than enforce it this way because I've been bitten by it before.

karenetheridge: Moo::sification is global, but it only prevents Moo->Moose inflation. It should not prevent other Moose-using classes from being loaded.

perigrin: This is why I didn’t want to block the merge. In the case of sqitch we were specifically trying to subclass a Moo class with a Moose class, or in a class where Moose had been loaded first. Moo saw Moose was loaded and so would attempt to load the Moose metaclass and `Moo::sification` would then blow up.

I don’t know that a similar situation would arise here, so I don’t wanna block the merge because of that. I’m jut giving my two cents.

karenetheridge:
perigrin approved this before the most recent rebase, so I'm considering this to be two reviews, so merging.

PR #646, opened 2019-02-06
fix workspace-rack query

### LINKED ISSUE #643: Unexpected racks appearing in workspaces
When a client issues a `GET /workspace/:wid/rack` request, the API returns devices that don't have a direct relationship with that workspace. A subsequent to `GET /workspace/:wid/rack/:rid` where :rid is the UUID for one of these racks returns a 404.

Workspaces should only display the racks they have a direct relationship with (either because they have a common row in workspace_datacenter_rack or because the rack has a room_id that has a common row with the workspace in workspace_datacenter_room.

If a rack does appear in the results from `GET /workspace/:wid/rack` then it SHOULD always return valid results for `GET /workspace/:wid/rack/:rid` and not a 404.

karenetheridge:
Instead of only fetching racks in this workspace, we were also fetching all sibling racks (via
the workspace-racks' parent rooms) as well.
    
fixes #643.

### DIFF
--- a/lib/Conch/Controller/WorkspaceRack.pm
+++ b/lib/Conch/Controller/WorkspaceRack.pm
@@ -30,7 +30,7 @@ sub list ($c) {
     my $racks_rs = $c->stash('workspace_rs')->associated_racks->active;
 
     my $device_health_rs = $racks_rs->search(
-        {},
+        { 'device.id' => { '!=' => undef } },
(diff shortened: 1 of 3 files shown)


karenetheridge:
backported and deployed in 2.23.2 (staging) and 2.22.6 (production).

Issue #625, opened 2019-01-25
Commands for initializing and migrating the database


sungo:
I'm looking at creating docker compose files to standing up conch and its dependencies. Typically, one uses the official postgres image and points the application to it. The implication here is that the official image doesn't have any of our sql files or scripts on it.

When standing up conch for the first time, the database needs to be initialized. Later, as the user upgrades the image, the database needs to be migrated. Since the API does not, for good reason, manage these tasks on its own, I need a way to perform these tasks out of band.

I'd like to have mojo commands, that use the regular conch/mojo config system, to initialize and migrate the database. Example syntax might be `docker run --rm joyentbuildsops/conch-api bin/conch init-database` and `docker run --rm joyentbuildops/conch-api bin/conch migrate-database`. 

The conch docker image contains ubuntu's build-essential package so these commands could just be `system` commands but I leave that implementation detail up to the implementer.

sungo:
No. The code will not know that it is running inside docker. This is just like standing up and migrating databases in production and doesn't require any extra considerations.

karenetheridge:
This should be really easy after the (planned and almost done, via validation refactoring) database renovations are in, as I've already done nearly the same thing in Test::Conch for initializing a fresh db for tests.  The only caveat is that when getting `$self->app` in a command, the app is already initialized and we will pass through the "make sure the database has all its migrations up to date" check, which we'll need to bypass/defer with some sort of special hook, hopefully one not too hacky.

karenetheridge:
What about populating the db with basic data in order to run the application? Is that out of scope for this ticket?  Normally we expect:

- a user with system_admin privileges that can create other users
- a `workspace` named 'GLOBAL'
- one or more `validation_plan` entries with `validation`s
- probably more that I'm not thinking of presently

sungo:
I don't know what basic data is required for conch to work initially. Accepting auth information for a first admin user makes sense. If GLOBAL is required for the API to operate out of the box, then yes create that too. If additional basics are required, then either the script needs to provide those or we need to examine why they're necessary for an empty instance.

The goal here is pretty simple. Given a postgres database and a conch docker image, I want to stand up a new instance of the conch API.  When upgrading that instance later, I want to migrate the database so it keeps working.

The command syntax you suggested is fine.

karenetheridge:
> Given a postgres database 

[...]

PR #656, opened 2019-02-15
Migrate to Sqitch

perigrin:
This replaces our homegrown migration system with [App::Sqitch](https://metacpan.org/pod/App::Sqitch). 

It also adds a basic command system for manipulating the database from the command line using  `conch.conf`.


misc/move_migrations_to_sqitch.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
karenetheridge: This file looks like something you ran to create the new files - does it need to be kept around?

sql/deploy/add-device-triton-setup.sql
```diff
@@ -1,5 +1,5 @@
-SELECT run_migration(18, $$
+BEGIN;
```
karenetheridge: Since I doubt these old migrations will even run anymore (and some do naughty things with data presuming the presence of old data that won't exist on a fresh db), do we need to keep most of these files at all? We could pick a line in the sand like the 2.24 release and forget about all migrations before that point, and start with that release's sql/schema.sql instead.  Migrations after that point will need to be written with sqitch in mind in case there are any other things to consider (are there?)

Makefile
```diff
@@ -50,7 +50,7 @@ migrate-db: ## Apply database migrations
 
 .PHONY: db-schema
 db-schema: ## create a dump of current db schema
-	pg_dump --username conch --schema-only --file sql/schema.sql conch
+	@carton exec bin/conch_db dump-schema --file sql/schema.sql
```
karenetheridge: Heh, I also created a bin/conch_db executable in my branch for #625 (it's much shorter, doing a few utility things, some of which overlap here) :)

sql/deploy/add-device-triton-setup.sql
perigrin: I'm fine with that idea but I didn't want to bite *that* discussion off in what is mostly a simple transition of underlying tooling. I'd rather do that in a separate PR.

Makefile
perigrin: Yep, I specifically used #625 as the basis for this command, the WIP tag is mostly because I don't have all of the things you do there implemented yet.

misc/move_migrations_to_sqitch.sh
perigrin: It doesn't, I'll kill it.

perigrin: killed.

sql/deploy/add-device-triton-setup.sql
perigrin: So what I've done (and I'm happy to re-visit this if you think it's warranted) is I've put as my last commit a migration that purges the old migrations table (if it exists). The rollback for this migration throws an exception and stops everything before it from happening.

I've done some minimal work to make sure you _could_ rollback before this commit but I haven't gone out of my way to make sure that it leaves the application in a sane state. Simply that it attempts to roll things back such that you can process reversions before that relatively cleanly. (For example I don't attempt to add-back dropped columns, or drop new columns, I don't attempt to add back all the purged Workflow tables etc.)

This means that it the history of these changes is minimally preserved in Sqitch, and if necessary we _can_ re-visit these migrations (rework them in sqitch terms etc). But by default they're gated behind a line in the sand.

Makefile
perigrin: Yours actually was `bin/conch-db` and so there wasn't a conflict, I just had to port my stuff over. I've since abandoned my own `bin/conch_db` because yours was less invasive.

karenetheridge:
Shall we try again for v3?

sungo:
The concerns I've shared in chat are unlikely to change for v3. I'm not convinced we need/want the features provided by skitch and I think it adds unneeded complexity.

karenetheridge:
One of the features that sqitch has that would greatly help us is the ability to test (via the 'assert' file) if a schema change has been applied, and if it hasn't, to do so -- rather than relying on numeric entries in a 'migration' table. I frequently have to manually hack my local database when switching between branches in order to apply/unapply various migrations so as to match the local code.  Also the ability to roll back (while it may have some limitations) would be of great help in production.

perigrin:
Yeah basically I was conceding this *specific* PR wasn't the right answer.

Issue #618, opened 2019-01-18
allow overriding database credentials from the environment


karenetheridge:
e.g. docker may provide its creds through the environment: https://hub.docker.com/_/postgres

(this also needs a corresponding change in buildops-infra).

PR #659, opened 2019-02-19
command-line functionality to initialize the database and run migrations

### LINKED ISSUE #625: Commands for initializing and migrating the database

karenetheridge:
closes #625.

### DIFF
--- a/bin/conch-db
+++ b/bin/conch-db
@@ -1,0 +1,158 @@
+#!/usr/bin/env perl
+
+=pod
+
+=head1 NAME
+
+conch-db - commands to administrate a database
+
+=head1 SYNOPSIS
+
+    bin/conch-db [long options...]
+
+=head1 DESCRIPTION
+
+Work with the Conch database.  Run C<bin/conch-db --help> for a list of options.
+
+=cut
+
+use v5.26;
+use warnings;
+use utf8;
+use experimental 'signatures';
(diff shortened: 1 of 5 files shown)


karenetheridge:
Should POSTGRES_DB and POSTGRES_HOST be supported by the api itself also?  as in #633 and #618, currently we only support POSTGRES_USER and POSTGRES_PASSWORD there.

sungo:
@karenetheridge yes

karenetheridge:
I've updated the PR to also use POSTGRES_DB and POSTGRES_HOST, in both this new script and the api in general.

PR #662, opened 2019-02-21
new /datacenter_rack/:id/assignment endpoints, for device-rack associations

### LINKED ISSUE #654: New endpoints for assigning devices to racks
Currently, devices are assigned to racks using `/workspace/:id/rack/:id/layout`. It accepts a payload that looks like `{ "device_id": 47 }`

The current endpoint has a few problems that I'd like to resolve:
* Requires a workspace 
* The payload is unfortunate
* The UI now offers the ability to update a device's asset tag during this stage

To resolve these issues, I'd like:
* Use new endpoints - this prevents breaking the web UI
* Not require a workspace - Since the user may have the ability to write to a rack via several workspaces, I'd like the API to resolve those permissions internally, rather than have to figure out an appropriate workspace for every operation
* Allow setting a device's asset tag in the same operation as assigning it to a rack slot
* Use a new payload which I will detail below.

I suggest:
* `GET /rack/:id/assignment` - returns the current rack configuration
* `POST /rack/:id/assignment` - updates the current rack configuration, assigning devices as necessary
* `DELETE /rack/:id/assignment` - deletes devices from rack slots

The GET payload would look like: 
```
[
  {
    "rack_unit_start": 47,
    "device_id": "SN1234",
    "device_asset_tag": "JOY1234",
    "hardware_product": "friendly-name",
  },
  {
    "rack_unit_start": 45,
    "hardware_product": "friendly-name",
  },
]
```

The POST payload would look like:
```
[ 
  { 
    "device_id": "SN1234",
    "rack_unit_start": 47,
    "device_asset_tag": "JOY1234",
  },
  {
    "device_id": "SN1235",
    "rack_unit_start": 45,
    "device_asset_tag": "JOY1235",
  },
]
```
* asset tags should be optional

The DELETE payload would look like:
```
[
  {
    "rack_unit_start": 47,
    "device_id": "SN1234",
  },
]
```

karenetheridge:
closes #654.

Also adds workspace permission checks for `/rack/*` endpoints.

### DIFF
--- a/json-schema/response.yaml
+++ b/json-schema/response.yaml
@@ -1343,6 +1343,38 @@ definitions:
       updated:
         type: string
         format: date-time
+  RackAssignments:
+    type: array
+    uniqueItems: true
+    items:
+      $ref: /definitions/RackAssignment
+  RackAssignment:
+    type: object
+    additionalProperties: false
+    required:
+      - device_id
+      - device_asset_tag
+      - hardware_product
+      - rack_unit_start
+      - rack_unit_size
+    properties:
+      device_id:
+        oneOf:
+          - $ref: /definitions/device_id
+          - type: 'null'
+      device_asset_tag:
+        oneOf:
+          - $ref: /definitions/device_asset_tag
+          - type: 'null'
+      hardware_product:
+        type: string
+        description: name of the hardware_product designated for that slot
+      rack_unit_start:
+        type: integer
+        description: Starting RU slot position
+      rack_unit_size:
+        type: integer
+        description: number of slots (expected to be) occupied by the hardware
   User:
     type: object
     additionalProperties: false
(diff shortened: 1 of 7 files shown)


json-schema/response.yaml
```diff
@@ -1343,6 +1343,38 @@ definitions:
...
+        oneOf:
+          - $ref: /definitions/device_asset_tag
+          - type: 'null'
+      hardware_product:
+        type: string
+        description: name of the hardware_product designated for that slot
+      rack_unit_start:
+        type: integer
+        description: Starting RU slot position
+      rack_unit:
```
sungo: This name is confusing and I'd prefer it be something like "rack_unit_size"

karenetheridge: done.

PR #667, opened 2019-02-26
fixes for 2.24

karenetheridge:
- fix a bad device query
- properly update the `updated` column when mutating various database data
- when creating a new user, move transmission of the optional `is_admin` flag's value out of the URL into the payload body where it belongs (this affects conch-shell v2)
- further improvements to json schemas and data validation

 The first commit, fixing the `/device/:id/*` endpoints' query, should be backported into 2.23 (as that is the version in which the poor performance was introduced).


lib/Conch/Controller/User.pm
```diff
@@ -363,10 +363,9 @@ sub list ($c) {
 
 Creates a user. System admin only.
 
-Optionally takes a query parameters:
+Optionally takes a query parameter:
 
 * 'send_mail' (defaulting to true), to send an email to the user with the new password
```
sungo: Should send_mail be pulled into the post body as well?

karenetheridge: IMO no, because it's not a value being set in the db but rather an out of band thing to signal something else to do at the same time. but I'll defer to our REST expert on this :)

perigrin: Since this is a POST only endpoint it REST doesn't have an opinion here. Either is acceptable.

PR #669, opened 2019-02-26
populate validation_state rows from old device_validate backup

### LINKED ISSUE #660: Forge validation states for old device reports
With the new retention policy, we're looking to purge old records that don't meet the new "keep" requirements. For records that predate the current validation system, this is difficult since they lack validation state records. We don't know which are passing, failing, etc. We can't run them back through the current validation system as backwards compatibility has been broken several times since then.

It should be possible, however, to forge validation state records using the old device_validate table. We purged that table out of production back in August but the data still exists as a backup in manta.

I recommend we load that backup into a dev instance someplace and use the data to create validation states for all the old device reports, importing those states back into production. I'd like to keep the original date if we can, rather than creating them with $today's date.

karenetheridge:
This is to be run ahead of bin/conch thin_device_reports, which backfills results from archived
device_validate rows into validation_state.  Only overall error/fail/pass results are stored,
not the reasons for the results (which can be determined by looking at the individual
device_validate.validation json blobs).  All results for a given device_report_id are combined
together into a single validation_state entry, and are linked to a new validation_plan_id
created expressly for this script.

The input CSV file is a file prepared offline from a device_validate backup and shall be made
available on the target system(s) for running after deployment.

closes #660.

### DIFF
--- a/lib/Conch/Command/insert_validation_states.pm
+++ b/lib/Conch/Command/insert_validation_states.pm
@@ -1,0 +1,153 @@
+package Conch::Command::insert_validation_states;
+
+=pod
+
+=head1 NAME
+
+insert_validation_states - insert new validation_state records from old device_validate data
+
+=head1 SYNOPSIS
+
+    bin/conch insert_validation_states [long options...] inputfile.csv
+
+        --help  print usage message and exit
+
+=cut
+
+use Mojo::Base 'Mojolicious::Command', -signatures;
+use Getopt::Long::Descriptive;
+use Text::CSV_XS;
+
+has description => 'Insert new validation_state records from old device_validate data';
+
+has usage => sub { shift->extract_usage };  # extracts from SYNOPSIS
+
+sub run ($self, @opts) {
+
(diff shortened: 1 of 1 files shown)


lib/Conch/Command/insert_validation_states.pm
```diff
@@ -0,0 +1,153 @@
...
+
+        my $device_report_rs = $schema->resultset('device_report');
+        my $validation_state_rs = $schema->resultset('validation_state');
+
+        my $rows_created = 0;
+        my %exists_report_ids;
+        my %not_exists_report_ids;
+
+        my $validation_plan = $schema->resultset('validation_plan')->update_or_create({
+            name => 'Conch v1 placeholder',
```
sungo: I suggest making the name here something a bit more clear like 'device_validate placeholder' since we cut v2 when the mojo-ization happened, not when we changed validation systems.

Not a show stopper though.

karenetheridge: ok, will do that now

PR #678, opened 2019-03-02
make datacenter_room.alias not-nullable

### LINKED ISSUE #676: Make datacenter_room.alias not null
Ops is looking to use datacenter_room.alias to input the name for each room used by netops and dcops.  As such, it's important that this field become not-null to ensure that automation and humans can keep track of the rooms. 

Looking at production data, it seems all rows have this field already.

I'd like for this constraint to be backported to 2.24.

karenetheridge:
closes #676.

### DIFF
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -65,6 +65,7 @@ definitions:
     required:
       - datacenter
       - az
+      - alias
     properties:
       datacenter:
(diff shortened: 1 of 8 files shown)


karenetheridge:
I neglected to cherry-pick this into 2.24.3 for a staging release, but can do so now/this week for 2.24.6 for production.

PR #696, opened 2019-03-13
allow system admins to access all devices

### LINKED ISSUE #695: Ability to check for the existence of any device, regardless of location
Right now, there appears to be no way to get an HTTP 200 back from `GET /device/:id` for any device in the database. The device must be located to a workspace that the user can see or reported using a relay associated with the user's account. I need a way for some user type, probably admin sadly, to `GET /device/:id` for any device in the database, regardless of its location or workspace assignment.

karenetheridge:
closes #695.

### DIFF
--- a/lib/Conch/Controller/Device.pm
+++ b/lib/Conch/Controller/Device.pm
@@ -49,7 +49,10 @@ sub find_device ($c) {
         return $c->status(404, { error => 'Not found' });
     }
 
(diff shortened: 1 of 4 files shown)


karenetheridge:
This will be backported to 2.24 (production).

karenetheridge:
I have backported this to release/v2.25 and tagged it as 2.25.1.

PR #702, opened 2019-03-19
Support docker-compose

sungo:


Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apt-get update && \
 		git \
 		libssl-dev \
 		libzip-dev \
+		perl-doc \
```
karenetheridge: `carton exec -- bin/conch-db --help` already worked though, right?

conch.conf.dist
```diff
@@ -36,6 +36,10 @@
 		audit          => 0,
 		rollbar        => 0,
 		nytprof        => 0,
+
+		# For docker, this should probably be set to 1. If not, logs will be
+		# written to /app/conch/log inside the container and a volume will need
+		# to be mounted there.
```
karenetheridge: There is another option: set `logging => { dir => $dir }` (defined lower in the file) to some filesystem that is mounted.

Dockerfile
sungo: Yes

conch.conf.dist
sungo: The documentation remains the same, whether the app is logging to `/app/conch/log` or `/var/log`. No directories are mounted by default. If the user does not set `log_to_stderr`, the API will attempt to write logs somewhere on disk that will remain inaccessible unless they mount a volume on that place.

karenetheridge: "this should probably be set to 1" is not quite right though, since there is another option.

sungo: Docker, in its default configuration, grabs stdout and stderr from the application and logs it somewhere based on the docker daemon's configuration indicates. For example, in our edge environment, docker writes the logs to syslog.

If the user sets `log_to_stderr` to 1 in the config, which is the comment indicated here, the API logs to stderr which will either be shown on the cli directly or in docker's regular logs. This is the standard operating practice for docker.

if the user sets `log_to_stderr` to 0, by default, the logs will be written to `/app/conch/log` and the user must mount a volume to catch those. This requires extra steps and deviates from the usual docker practices. So it "probably" should be set to 1 so those behaviors are maintained.

karenetheridge: ok.

PR #723, opened 2019-04-02
refactor json schema handling

karenetheridge:
rewrite code to extract schema definitions, adding tests and fixing many bugs:

- find and traverse $refs within $refs
- handle recursive $refs
- support references to other files or URLs
- inline $refs at the top level of the requested definition
- do not corrupt the schema in the validation object, which is used across
  many requests
- detect unsupported namespace collisions (would need a renaming heuristic)

factor out common definitions into a separate schema file

- added a version to the incoming DeviceReport
- device reports in responses are now looser, as they contain data in a
  variety of historical input formats

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -15,7 +15,7 @@ requires 'Mail::Sendmail';
 requires 'Try::Tiny';
 requires 'Time::HiRes';
 requires 'Time::Moment', '>= 0.43'; # for PR#28, fixes use of stdbool.h (thanks Dale)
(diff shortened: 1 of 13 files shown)


karenetheridge:
I've pulled the device report schemas (just one for now, DeviceReport_v2.24) into its own file.  This exposed a new edge case in the schema ref resolution code which I've now fixed (and added a few more test cases to capture these edge cases). The implementation is a tad bit simpler now too as a result, so yay.

PR #725, opened 2019-04-04
add "phase" column to device

### LINKED ISSUE #621: Phases
Currently Conch only validates devices during "preflight" and has no concept of 
Phases. We are planning on adding in-flight reports from production machines
and the reality of pre-flight is that it has two distinct phases. So we will
need to introduce the concept of phases to conch reports. Specifically for 
our current preflight uses we will need:

1. `Integration` 
2. `Installation`
3. `Production`

Integration is the assembly of the racks at the integration facility. 
Installation is the delivery and installation of the racks at the DC, and
Production is the final sign-off and accepting user requests.

Eventually we will need to divide Production into `Production` and 
`Diagnostic/Repair` to allow for in-production downtime for operations
(hardware swaps, firmware updates etc). We may also at some point add 
post-production phases like `RMA`, `End-of-Life` and `Decommissioned`.

Phases will be set via a new endpoint `/device/:id/phase` (TBD) similar 
to how device locations are set. Until we start tracking in-production 
and _especially_ `Diagnostic/Repair` states we can assume that a given 
report is for `Integration` unless told otherwise.

With the addition of phases we will need to update the report retention policy 
to retain the first and last reports for a given phase.

karenetheridge:
new endpoints:
    GET /device/:id/phase
    POST /device/:id/phase

the default phase for all existing and new devices is "integration".

closes #621.

### DIFF
--- a/json-schema/common.yaml
+++ b/json-schema/common.yaml
@@ -22,6 +22,13 @@ definitions:
   device_asset_tag:
     type: string
     pattern: ^\S+$
+  device_phase:
+    type: string
+    enum:
+      - integration
+      - production
+      - diagnostics
+      - decommissioned
   int_or_stringy_int:
     description: an integer that may be presented as a json string
     # note that when JSON::Validator has 'coerce' mode on, both of these rules will match.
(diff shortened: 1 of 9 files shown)


sungo:
I've added a ticket ( joyent/conch-shell#282 ) to track the required changes in the shell

PR #727, opened 2019-04-05
New endpoints to allow the user to create new JWTs for future authentication.

### LINKED ISSUE #726: API Token System for v2
Supersedes #478 

I would like an auth system by which a "user" (likely automation or the conch shell) can obtain a long-lived/permanent "token" and authenticate against the API using that "token".

 The high-level workflow I envision is: 
* user logs onto web ui
* user navigates to token section 
* user generates token 
* user pastes token to the shell configuration 
* user enjoys a happy life. 

 ## Architecture

* CRUD endpoints to manage tokens
* Auth system must accept these tokens in place of a user's regular login
* These tokens confer the same privileges as the user's regular login

## Microarchitecture

The API currently has the ability to generate JWT with different expiration dates. Admins can have one and regular users can have another.

I propose we introduce a third type that default expires to 5 years. This should be configurable in the config file, just as the timeouts for the other JWT classifications. 

The API needs a CRUD interface around generating those long-lived JWT.

From the client-side, the user passes up the "token" aka "long-lived JWT" in the Authorization header just like one does today.. 

This should prevent any significant changes to the auth system since it is accepting JWTs as it does today, just with a different expiration date.

### Nice To Haves

* A `name` for each "token" would allow the user to better audit their tokens
* A `last used` for each "token" would similarly allow a user to audit their tokens

karenetheridge:
closes #726.

### DIFF
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -442,6 +442,99 @@ sub deactivate ($c) {
 	return $c->status(204);
 }
 
+=head2 get_tokens
+
+Get a list of unexpired user tokens.
+
+Response uses the UserTokens json schema.
+
+=cut
+
+sub get_tokens ($c) {
+    my $rs = $c->db_user_session_tokens->unexpired->order_by('name');
+    return $c->status(200, [ $rs->all ]);
+}
+
+=head2 create_token
+
+Create a new token, creating a JWT from it.  Response uses the NewUserToken json schema.
+
+=cut
+
+sub create_token ($c) {
+    my $input = $c->validate_input('NewUserToken');
+    return if not $input;
+
+    return $c->status(400, { error => 'name "'.$input->{name}.'" is already in use' })
+        if $c->db_user_session_tokens
+            ->search({ user_id => $c->stash('user_id'), name => $input->{name} })->exists;
+
+    # default expiration: 5 years
+    my $expires_abs = time + (($c->config('jwt') || {})->{custom_token_expiry} // 86400*365*5);
+
+    # TODO: ew ew ew, some duplication with Conch::Controller::Login::_create_jwt.
+    my ($new_db_row, $token) = $c->db_user_session_tokens->generate_for_user(
+        $c->stash('user_id'), $expires_abs, $input->{name});
+
+    my $jwt = Mojo::JWT->new(
+        claims => { uid => $c->stash('user_id'), jti => $token },
+        secret => $c->config('secrets')->[0],
+        expires => $expires_abs,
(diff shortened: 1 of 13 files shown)


karenetheridge:
added an extra `delete from user_session_token where expires < now();` to the migration script to make it run faster.

lib/Conch/Controller/User.pm
```diff
@@ -442,6 +442,100 @@ sub deactivate ($c) {
...
+    my $jwt = Mojo::JWT->new(
+        claims => { uid => $c->stash('user_id'), jti => $token },
+        secret => $c->config('secrets')->[0],
+        expires => $expires_abs,
+    )->encode;
+
+    return $c->status(201, {
+        token => $jwt,
+        $new_db_row->TO_JSON->%*,
+    });
```
perigrin: I would also add a `Location` header to point to the endpoint you can GET the new db row data from.

EDIT: WIth a 201 that _shouldn't_ cause a client to re-direct automatically, but it may be that the world conspires against correctness here.

lib/Conch/Controller/User.pm
```diff
@@ -442,6 +442,100 @@ sub deactivate ($c) {
...
+database, stashing a resultset to access it as 'token_rs'.
+
+Only the current user may access the token.
+
+=cut
+
+sub find_token ($c) {
+    my $token_rs = $c->db_user_session_tokens->unexpired
+        ->search({ user_id => $c->stash('user_id'), name => $c->stash('token_name') });
+
```
perigrin: Does this mean we can list expired token rows but we can't look them up individually?

karenetheridge: yes. I can add `?with_expired=1` support to the individual GET request as well.

Note that in practice this will be pretty pointless though, as (as noted previously) we automatically delete all expired tokens when passing through the login code.

perigrin: Yeah. I don't like having one endpoint say "here's this data" and another endpoint say "what data? 404!". You're right that we can't _avoid_ that if someone issues fetches the list of expired tokens and then we delete them before they can fetch a single token ... but we can at least try to minimize it.

sungo: Let's look at it from a different direction. Why are we allowing users to access expired tokens at all, particularly if at some random point the code is going to delete them out from under us?

karenetheridge:
Updated to resolve last review comments -- can now fetch individual expired tokens; added Location header to 201 response upon creation.

karenetheridge:
Removed all uses of `?with_expires=..`.

PR #734, opened 2019-04-09
improvements for 2.27

karenetheridge:
Various improvements to the code, most of which have no external effects whatsoever.

- add missing documentation on new phase endpoints, remove outdated documentation, clarify when endpoints only return active resources
- remove useless tests that rely on there *not* being data
- folded uuid validation in endpoint paths into a route type method
- use consistent style in route definitions
- added some input validations where they were lacking
- small fixes in test subs
- include rollbar uuid in log messages, for better data correlation


karenetheridge:
> [2f481b0](https://github.com/joyent/conch/commit/2f481b0cb2fa14a44412a52dcfa1a5affc08d0c3) is unfortunate. Making sure that the system operates in the lack of data is critical to making sure a fresh installation works.

Not really. Most of the tests identical with tests in other parts of the system. There were two that were not, which I added to the proper location, and added a few more tests for error cases that never had any representation.

PR #737, opened 2019-04-15
sysadmin token endpoints

### LINKED ISSUE #735: Allow an admin to delete a user's named token
The API contains the ability for a user to delete all of their tokens or a token by name with an admin having permissions to delete all of a user's tokens.

Unless I'm misunderstanding, admins do not have the ability to delete a single token for a user by the token's name. I would like for admins to have that ability.

karenetheridge:
- fixed naming of login tokens
- new endpoints:

    DELETE /user/me/token
    GET    /user/#target_user_id_or_email/token
    DELETE /user/#target_user_id_or_email/token
    GET    /user/#target_user_id_or_email/token/:token_name
    DELETE /user/#target_user_id_or_email/token/:token_name

Token removal endpoints accept `?login_only=1`, `?api_only=1`

closes #735, #738.

### DIFF
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -539,6 +539,8 @@ Deactivates a token from future use.
 =cut
 
 sub expire_token ($c) {
+    $c->log->warn('user '.$c->stash('user')->name.' expired user session token "'
+        .$c->stash('token_name').'" for user '.$c->stash('target_user')->name);
     $c->stash('token_rs')->expire;
     return $c->status(204);
 }
--- a/lib/Conch/Controller/User.pm
+++ b/lib/Conch/Controller/User.pm
@@ -21,19 +21,37 @@ Conch::Controller::User
 
 =head2 revoke_user_tokens
 
-Revoke *all* of a specified user's session tokens and prevents future session authentication,
-forcing the user to /login again.
+Revoke a specified user's tokens and prevents future token authentication,
+forcing the user to /login again. By default *all* of a user's tokens are deleted,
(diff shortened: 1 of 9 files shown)


lib/Conch/Controller/User.pm
```diff
@@ -19,31 +19,36 @@ Conch::Controller::User
...
-
-Revoke *all* of a specified user's session tokens. System admin only.
+System admin only (unless reached via /user/me).
 
 =cut
 
 sub revoke_user_tokens ($c) {
 	my $user = $c->stash('target_user');
 
+    return $c->status(400, { error => 'cannot use login_only and api_only together' })
```
sungo: The default behavior is equivalent  to `login_only=1` combined with `api_only=1`, yes? So while it's silly to provide them both, why bother throwing an error?

karenetheridge: No, combining "login only" AND "api only" is the empty set, as would be `?login_only=1&api_only=1`. The default is `login_only=0&api_only=0`.

karenetheridge:
I have pushed some changes in response to #738. (compare to the previously-reviewed changes at commit 81ad20c6a2.)

karenetheridge:
> Changing one's password via POST /user/me/password should probably allow the option to clear one's API tokens

That already exists: https://github.com/joyent/conch/blob/master/lib/Conch/Controller/User.pm#L186-L191 and https://github.com/joyent/conch/blob/master/lib/Conch/Controller/User.pm#L234-L239

karenetheridge:
replaced by #743.

PR #742, opened 2019-04-18
add the concept of phases at the rack level

### LINKED ISSUE #741: Rack Phase API
After announcing the device phasing work done in                                                                                                                                                         #725 @jemershaw asked if we could set the phases                                                                                                                                              at the rack level so we could hook the phase changes into the existing tooling                                                                                                                              we do for promoting a rack to production. This makes total sense, and I'm sorry                                                                                                                             it was forgotten in the initial ticket.                                                                                                                                                                     
                                         
Also after discussion with @sungo I was convinced we need to bite the bullet and track the phase of the *rack* itself as well. Racks have device-like behavior (most importantly they're physical things in the world) that regardless of how we currently track them in the database need to be modeled in the API.
                                         
@karenetheridge suggested the following API and I think it makes sense (with some modifications):                                                                                                                                                
                                                                                                                                                                                                            
> `POST /rack/:id/phase?rack_only=<0|1>`                                                                                                                                                                                       
>                                                                                                                                                                                                            
> This will set the phase for the rack all devices currently in the rack layout. If the `rack_only` query param is set to `1|true` then *only* the phase for the rack is adjusted.

The phase information should be returned in all the places we currently return rack and device data. I think we can get this for free by simply adding a `phase` column to the `rack` table, and returning it along with the rest of the rack data. Also as pointed out in a chat with @karenetheridge we will get the device level `phase` information by simply making sure it's reported wherever we return DetailedDevice data.

karenetheridge:
new endpoint: POST /rack/:id/phase

which updates the phase of all devices located in the rack, and/or the phase
of the rack itself.

closes #741.

### DIFF
--- a/lib/Conch/Controller/Rack.pm
+++ b/lib/Conch/Controller/Rack.pm
@@ -344,6 +344,34 @@ sub delete_assignment ($c) {
     return $c->status(204);
 }
 
+=head2 set_phase
+
+Updates the phase of this rack, and optionally all devices located in this rack.
+
+Use the C<rack_only> query parameter to specify whether to only update the rack's phase, or all
+located devices' phases as well.
+
+=cut
+
+sub set_phase ($c) {
+    my $input = $c->validate_input('RackPhase');
+    return if not $input;
+
+    $c->stash('rack_rs')->update({ phase => $input->{phase}, updated => \'now()' });
+    $c->log->debug('set the phase for rack '.$c->stash('rack_id').' to '.$input->{phase});
+
+    if (not $c->req->query_params->param('rack_only') // 0) {
+        $c->stash('rack_rs')
+            ->related_resultset('device_locations')
+            ->related_resultset('device')
+            ->update({ phase => $input->{phase}, updated => \'now()' });
+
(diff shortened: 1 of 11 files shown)


lib/Conch/Controller/Rack.pm
```diff
@@ -344,6 +344,34 @@ sub delete_assignment ($c) {
...
+=cut
+
+sub set_phase ($c) {
+    my $input = $c->validate_input('RackPhase');
+    return if not $input;
+
+    $c->stash('rack_rs')->update({ phase => $input->{phase}, updated => \'now()' });
+    $c->log->debug('set the phase for rack '.$c->stash('rack_id').' to '.$input->{phase});
+
+    if (not $c->req->query_params->param('rack_only') // 0) {
```
perigrin: The `clear_tokens` param also allows `no` and `false` as false values. Should we be consistent between these two endpoints?

karenetheridge: they're already inconsistent: one is a bool and the other isn't.

PR #750, opened 2019-04-24
big cleanup

karenetheridge:
a bunch of stuff that has been piling up. NO functionality has changed, anywhere, at all.


lib/Conch/Controller/HardwareProduct.pm
```diff
@@ -45,9 +45,9 @@ sub find_hardware_product ($c) {
     # route restricts key to: sku, name, alias
     if (my $key = $c->stash('hardware_product_key')
             and my $value = $c->stash('hardware_product_value')) {
-        $c->log->debug("Looking up a HardwareProduct by identifier ($key = $value)");
+        $c->log->debug('Looking up a HardwareProduct by identifier ('.$key.' = '.$value.')');
```
perigrin: Places like this seem like a step backwards, we _lose_ clarify and simplicity by having to concatenate instead of interpolate the string.

lib/Conch/Controller/HardwareVendor.pm
```diff
@@ -78,8 +78,8 @@ sub create ($c) {
     return $c->status(403) unless $c->is_system_admin;
 
     if ($c->db_hardware_vendors->active->search({ name => $c->stash('hardware_vendor_name') }) > 0) {
-        $c->log->debug("Failed to create hardware vendor: unique constraint violation for name");
-        return $c->status(400, { error => "Unique constraint violated on 'name'" });
+        $c->log->debug('Failed to create hardware vendor: unique constraint violation for name');
+        return $c->status(400, { error => 'Unique constraint violated on \'name\'' });
```
perigrin: This is another place I feel like we lose some clarity of intent.

lib/Conch/Controller/Login.pm
```diff
@@ -65,7 +65,7 @@ sub _create_jwt ($c, $user_id, $expires_delta = undef) {
     );
 
     # this should be returned in the json payload under the 'jwt_token' key.
-    return "$header.$payload";
+    return $header.'.'.$payload;
```
perigrin: And things like this (and the one on line 132 below) also seem to go out of their way a bit.

lib/Conch/Controller/User.pm
```diff
@@ -95,7 +95,7 @@ sub set_setting ($c) {
 
     my $key = $c->stash('key');
     my $value = $input->{$key};
-    return $c->status(400, { error => "Setting key in request object must match name in the URL ('$key')" }) unless $value;
+    return $c->status(400, { error => 'Setting key in request object must match name in the URL (\''.$key.'\')' }) if not $value;
```
perigrin: I don't understand the intent of both of these changes, I've already brought up the quote change, but the `if not` seems a little pointless in this instance. Is there a performance difference?

lib/Conch.pm
```diff
@@ -96,11 +96,10 @@ sub startup {
...
             # HTTP status code 415 'Unsupported Media Type' is returned.
-            if ($headers->content_length) {
-                unless ($headers->content_type
-                    && $headers->content_type =~ /application\/json/i) {
-                    return $c->status(415);
-                }
+            if ($headers->content_length
+                and (not $headers->content_type
+                     or $headers->content_type !~ /application\/json/i)) {
+                return $c->status(415);
```
perigrin: So of the changes in this commit, this is the only one that honestly seems clearer to me than the original, and I'm not 100% convinced it is (speaking as someone trying to follow the change to make sure the logic is identical).

lib/Conch/Controller/HardwareProduct.pm
karenetheridge: I find string interpolation to be confusing because it is less explicit -- sometimes things get interpolated unintentionally and produces surprising effects.

lib/Conch/Controller/User.pm
karenetheridge: `unless` is a lot harder to read because one needs to push a stack frame "this entire expression is being negated". IMO.  I avoid it in new code and replace it when I can.

lib/Conch/Controller/HardwareProduct.pm
perigrin: I have no issue with only using double quotes when you want to explicitly interpolate. I find inline concatenation operators get in the way of the intent of  the string that's being logged or assigned or returned in the general case.

PR #752, opened 2019-04-29
Add new docs and autogeneration process for github pages based docs hosting

sungo:
To be clear, this is not intended to be complete documentation, but rather a base that we can start adding to.


docs/README
```diff
@@ -0,0 +1,6 @@
+Yeah, there's ruby stuff in here. You can ignore every bit of it unless you'd
```
karenetheridge: ruby???

sungo: please see the rest of the file :) jekyll is ruby. if you want to use jekyll to review the changes you've made, you can. or you don't have to.

sungo:
@karenetheridge No, it shouldn't. Jekyll / GHpages renders html versions of the markup. That link points to the html-ized version. if you link to the `.md` version, the user will get the markdown text file which is not what I want there.

karenetheridge:
Looks like this didn't incorporate the changes I made in my mail PR.

PR #765, opened 2019-05-08
seamless upgrading of validation plans, while remaining immutable

### LINKED ISSUE #546: Don't mutate existing validations and validation plans - create new ones!
Existing `validation_plan` and `validation` rows in the database should never be mutated -- or historical data (`validation_state`, linked to `device_report`) cannot be used to give an accurate picture of what checks were done on a device.

If a validation db row is found to not match a validation module on disk, deactivate it, remove it from all validation_plans, create a new validation row, deactivate all validation_plans containing that old validation, and create an equivalent one with the new validation as a member.

(Deactivating validation plans and creating new ones should be done very carefully, only with the explicit consent of the operator. Therefore only do this by execution of a command line utility, not as some side effect in the application itself.)

Builds on work done for #545.

karenetheridge:
closes #546.


lib/Conch/Validation/BiosFirmwareVersion.pm
```diff
@@ -3,12 +3,9 @@ package Conch::Validation::BiosFirmwareVersion;
 use Mojo::Base 'Conch::Validation';
 
 use constant name        => 'bios_firmware_version';
-use constant version     => 1;
+use constant version     => 2;
```
sungo: In some places in this commit, the version number of the validation is incremented, even though only the description is being updated. In others, the version number is *not* incremented. Seems like we might want a unified approach in the PR. I'm not dogmatic about which path.

karenetheridge: The version number is not incremented when it is not necessary to do so -- when the description has only changed by trimming whitespace off the start and end, which has already been reflected in the database.  The version is only incremented when the db record will change upon the next run of `Conch::ValidationSystem::load_validations`.

sungo: ah ok.

karenetheridge:
I pushed some minor updates to the last two commits, adding more tests.

PR #768, opened 2019-05-09
Drop /reset_password

### LINKED ISSUE #767: POST /reset_password needs attention
Currently the implementation is taking a POST request, validating the input arguments and then doing a redirect to `/user/#target_user_id_or_email/password` which only implements a `DELETE` method.  According to [rfc7321 section 6.4.2](https://tools.ietf.org/html/rfc7231#section-6.4.2) this isn't correct the only exception is:

> For historical reasons, a user agent MAY change the request
> method from POST to GET for the subsequent request.

It looks like we *should* be re-directing them to `/user/me/password` instead, and because we don't want them changing HTTP method we should be using `308 PERMANENT REDIRECT` (see [rfc7238](https://tools.ietf.org/html/rfc7238) instead.

Alternatively if we're nobody is using this endpoint anymore, we should just drop it to a `410 GONE`.

perigrin:
The previous implementation was broken and as far as I can tell totally unused. This closes #767.

### DIFF
--- a/lib/Conch/Controller/Login.pm
+++ b/lib/Conch/Controller/Login.pm
@@ -305,19 +305,6 @@ sub session_logout ($c) {
     $c->status(204);
 }
 
-=head2 reset_password
-
-Resets a user's password, given an email address, and sends the user an email
-with their new password.
-
-=cut
-
-sub reset_password ($c) {
-    my $input = $c->validate_input('ResetPassword');
-    return if not $input;
-    return $c->status(301, '/user/email='.$input->{email}.'/password');
-}
-
 =head2 refresh_token
 
 Refresh a user's JWT token. Deletes the old token.
--- a/json-schema/input.yaml
+++ b/json-schema/input.yaml
@@ -422,14 +422,6 @@ definitions:
           - $ref: /definitions/email_address
       password:
         $ref: /definitions/non_empty_string
-  ResetPassword:
-    type: object
-    additionalProperties: false
-    required:
-      - email
-    properties:
-      email:
-        $ref: /definitions/email_address
   UserPassword:
     type: object
     additionalProperties: false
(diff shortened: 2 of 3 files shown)


t/pod-spelling.t
```diff
@@ -55,6 +55,7 @@ mtu
 nullable
 ORed
 postgres
+PostgreSQL
```
karenetheridge: I don't see why this is necessary for this PR.

perigrin: It's not, it's a pre-existing failure that's causing `t/pod-spelling.t` to fail (something I have running now because of #766).

I'm happy to split it out into a separate PR if you'd like.

lib/Conch/Controller/Login.pm
```diff
@@ -305,19 +305,6 @@ sub session_logout ($c) {
...
 
-=head2 reset_password
-
-Resets a user's password, given an email address, and sends the user an email
-with their new password.
-
-=cut
-
-sub reset_password ($c) {
-    my $input = $c->validate_input('ResetPassword');
```
karenetheridge: Please also remove the json schema since it will no longer be used.

t/pod-spelling.t
karenetheridge: Yes, this is a separate issue that I'd like to track down and resolve separately. It's not failing on master, obviously, so there is something else going on.

perigrin: Dropped in the latest rebase.

lib/Conch/Controller/Login.pm
perigrin: So yeah my suggestion in the original ticket that we re-direct to `/user/me/password` was entirely wrong since the request bodies for those two endpoints are entirely different.

It's purged in the most recent rebase.

karenetheridge:
please remember to set milestones for all PRs.

PR #775, opened 2019-05-10
Add DeviceNicField

perigrin:
Needed to cover responses from `GET /device/:device_id/interface/:interface_name/:field`


perigrin:
POD for the controllers and the tests have been updated. `#get_all` was already covered by `DeviceNics` in the tests, I've added a second commit that documents all the return values for the rest of the methods in the Controller.

PR #782, opened 2019-05-16
Align the Workspace route with the WorkspaceDevice controller

perigrin:
A couple of our query parmas in `Conch::Controller::WorkspaceDevice`
don't explicitly need a `T` or `F`. This aligns the redirect and docs in
the route with the expectations in documented in the controller.


karenetheridge:
'parmas' in commit comment

perigrin:
Honestly looking at #781 and the underlying code. This was fixing something that isn't a problem anymore. I'm closing this. We can open a new PR if we want to add a test for the `active=1` state.

PR #786, opened 2019-05-17
validation for HDD sizes

### LINKED ISSUE #574: Want validation for hdd block size
On request of CPG, the new Ceres Type-2 and NCP Type-1 servers are being shipped with their hard drive block sizes set to 4k native rather than the default 512e (emulated). It has been requested that we verify the block sizes on these servers that they are indeed 4kn and raise a validation fail if any are found not to be. Corresponding issue for adding block size to the reporter is joyent/conch-livesys#112

karenetheridge:
This relies on the use of the 'specification' field in the hardware_product
table, specifically a section structured thusly (encoded as a json string):

    {
      disk_size => {
        <disk model> => <size in blocks>,
        ...,
        _default => <size to assume if the model is not listed here>,
      },
    }

Every device sku that needs a hard drive size validation will need this data
entered into its hardware_product.specification. Those skus that do not need
such validation need not have their data modified, and the validation will be
skipped.

closes #574.

Requires a utility to be run to add the validation to the appropriate plan before it can be used.

### DIFF
--- a/lib/Conch/Validation/HddSize.pm
+++ b/lib/Conch/Validation/HddSize.pm
@@ -1,0 +1,53 @@
+package Conch::Validation::HddSize;
+
+use Mojo::Base 'Conch::Validation', -signatures;
+use Mojo::JSON 'from_json';
+
+use constant name        => 'hdd_size';
+use constant version     => 1;
+use constant category    => 'DISK';
+use constant description => 'Check hard drive sizes';
+
+sub validate ($self, $data) {
+    my $hw_spec = from_json($self->hardware_product_specification // {});
+
+    return if not $hw_spec->{disk_size};
+
+    foreach my $disk_serial (keys $data->{disks}->%*) {
+        my $drive_model = $data->{disks}{$disk_serial}{model};
+        if (not $drive_model) {
+            $self->fail('missing drive model for disk '.$disk_serial,
+                component_id => $disk_serial);
+            next;
+        }
+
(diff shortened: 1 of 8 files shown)


lib/Conch/Validation/HddSize.pm
```diff
@@ -0,0 +1,53 @@
...
+
+sub validate ($self, $data) {
+    my $hw_spec = from_json($self->hardware_product_specification // {});
+
+    return if not $hw_spec->{disk_size};
+
+    foreach my $disk_serial (keys $data->{disks}->%*) {
+        my $drive_model = $data->{disks}{$disk_serial}{model};
+        if (not $drive_model) {
+            $self->fail('missing drive model for disk '.$disk_serial,
```
sungo: I think we should just skip the validation if we don't have data to compare, rather than fail

karenetheridge: Could do. I *think* model is a mandatory field however? (Not that our schema says it is, but my reading of the livesys code suggests that it's always there.)

sungo: If we can get confirmation of that and get it into the schema or at least documentation/comments, then I'm fine with failing. 

I should also have put a similar comment in the next `->fail` as well.

perigrin: If we have a disk_size in the hardware specification we are expecting to have that data in the report, if it’s not in the report something is wrong.

sungo: Understood.

PR #819, opened 2019-06-20
v3: small response changes

karenetheridge:
outward-facing changes:

- return a consistent 204 response from `POST /workspace/:workspace_id_or_name/user?send_mail=<1|0>` (instead of sometimes 200)
- when responses are lists, the entries are sorted sensibly
- `POST /user/:id_or_email` now redirects to its corresponding GET rather than returning the user directly, to be more consistent with other POST endpoints


docs/modules/Conch::Route::User.md
```diff
@@ -14,7 +14,7 @@ Unless otherwise noted, all routes require authentication.
 
 - Response: response.yaml#/UserDetailed
 
-### `POST /user/me/revoke?send_mail=<1|0>& login_only=<0|1> or ?api_only=<0|1>`
+### `POST /user/me/revoke?send_mail=<1|0>& login_only=<0|1> or &api_only=<0|1>`
```
perigrin: Probably should be `POST /user/me/revoke/?send_mail=<1|0>&login_only=<0|1>&api_only<0|1>` and not have the `or` in there. The text below covers the fact that they're mutually exclusive in a clearer way.

lib/Conch/Route/User.pm
```diff
@@ -136,7 +136,7 @@ Unless otherwise noted, all routes require authentication.
 
 =back
 
-=head3 C<< POST /user/me/revoke?send_mail=<1|0>& login_only=<0|1> or ?api_only=<0|1> >>
+=head3 C<< POST /user/me/revoke?send_mail=<1|0>& login_only=<0|1> or &api_only=<0|1> >>
```
perigrin: See my comment above for the other endpoint that has a similar URL.

perigrin:
I'm not sure how @sungo got removed as a reviewer. All I did was mark it approved because my comments weren't IMO blocking.

lib/Conch/Route/User.pm
karenetheridge: that's the same endpoint :)

perigrin: D'oh. Not enough sleep this week.

PR #820, opened 2019-06-21
v3: small db changes

karenetheridge:
Visible changes:
- 'processing' is no longer a possible value for validation_state.status and validation_result.status (and responses that reference these columns)
- validation_result.component_id is now known as component
- hardware_product_profile.rack_size has moved to hardware_product.rack_unit_size


karenetheridge:
There are other changes related to relays that should be made, so I removed it from this PR.

PR #838, opened 2019-07-23
In Docker, upgrade Postgres to v10

sungo:
Both dockerfiles now use postgres 10 which ships with ubuntu:bionic.

Part of https://github.com/joyent/buildops-infra/issues/87

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 local
+postgres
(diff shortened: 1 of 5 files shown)


sungo:
@karenetheridge upside of buildbot's setup is that buildbot just tested the dockerfile changes :) 

Odd that the version assertion didn't get triggered during tests.

karenetheridge:
The version check is just a minimum, and does not check a maximum. But once we're on 10 we can raise that up to 10.latest (currently 10.9 -- https://www.postgresql.org/about/news/1949/)

sungo:
LTS is shipping 10.9 as well https://packages.ubuntu.com/bionic/database/postgresql-10

sungo:
@karenetheridge I've updated the version check

PR #864, opened 2019-08-13
stop running migrations if one failed

karenetheridge:


sungo:
if you're using bash, `set -e` will stop the script if any command failed. It's a bit more idiomatic but ends up pretty similar to your patch.

PR #881, opened 2019-09-10
hub.docker is no longer supported

sungo:


karenetheridge:
Does this need merging to release/v2 also?

Issue #807, opened 2019-06-05
determine a device's intended hardware by the sku in the device report


karenetheridge:
For validation purposes, we look up a device's intended hardware using the device's current location (when it has one), rather than using information in the device report.  In v3 the location is purely informational (and may be inaccurate after preflight), so the report information should always be used as the source of truth.

#806 refines the hardware product lookup mechanism to make it more consistent.

karenetheridge:
This was implemented by #807. Placeholder skus were inserted for those hardware_product rows that lacked them.

Issue #805, opened 2019-06-05
populate all empty hardware_product.sku columns


karenetheridge:
We need to use the hardware sku to determine which validation(s) to run against the device report. This column is presently nullable, and not all rows have this value populated on production.  This is a blocker to future validation work.

PR #892, opened 2019-09-25
v3: use device report sku to determine intended hardware

### LINKED ISSUE #806: determine validations, validation_plan etc based on hardware sku
Presently the algorithm for determinine which validation_plan to run against a device, and the hardware_product that applies to the device (where some validation parameters are stored) is very hacky, based on the report's 'device_type' and 'product_name' fields.

reference: https://github.com/joyent/conch/blob/release/v2/lib/Conch/Controller/DeviceReport.pm#L461-L503 (`_get_hardware_product`, `_get_validation_plan`).

We should instead use the report's sku to look up the hardware_product and validation_plan.

Blocked by #805 as not all rows have the sku column populated on production.

### LINKED ISSUE #893: create device+sku in advance of livesys
As discussed:

- something (e.g. netbox, or DCOPS master spreadsheets) will call conch with a serial number + sku, and conch will record this (creating/updating a device in the database)
- livesys will query conch for the sku corresponding to a specific serial number so it can pick its firmware loadout
- when a device report is processed, the corresponding device will now always exist, and the sku must match what conch has on file (else signal an error indicating that the firmware loadout is wrong).

I propose:
* `POST /build/:build_id_or_name/device` - with the payload containing the serial number and sku (submitting multiple devices at once probably makes sense). the device is created or updated.
  * normal response: 204.
  * error responses: 401 if the user cannot authenticate. 403 if the user doesn't have permission for the endpoint (the 'rw' role for the build would make sense here).
* `GET /device/:serial/sku` - get the device's sku (and possibly other relevant device-level fields  (`GET /device/:serial` returns a lot of data, most of which is not relevant at this stage of the process)
  * normal response: 200 plus payload.
  * error responses: 401 if the user cannot authenticate. 403 if the user doesn't have permission for the endpoint (minimum 'ro' role for the build the device is in).

karenetheridge:
- hardware_product.sku is now required (existing entries that lack it get an obviously-placeholder value until that can be corrected - see #805).
- the sku in the incoming device report is now used as the sole mechanism to determine a device's hardware_product
- reject a device report if it references a device that is not yet known to conch
- new endpoints to match the new device lifecycle:
  - `POST /build/:id_or_name/device` (create device with sku as part of a build)
  - `GET /device/:id_or_serial/sku`

closes #893, #807.

### DIFF
--- a/docs/json-schema/device_report.json
+++ b/docs/json-schema/device_report.json
@@ -205,6 +205,9 @@
         "serial_number" : {
           "$ref" : "common.json#/definitions/device_serial_number"
         },
+        "sku" : {
+          "type" : "string"
+        },
         "system_uuid" : {
           "$ref" : "common.json#/definitions/non_zero_uuid"
         },
@@ -236,6 +239,7 @@
       "required" : [
         "bios_version",
         "product_name",
+        "sku",
         "serial_number",
         "system_uuid"
       ],
--- a/docs/json-schema/request.json
+++ b/docs/json-schema/request.json
@@ -315,6 +315,7 @@
             "name",
             "alias",
             "hardware_vendor_id",
+            "sku",
             "rack_unit_size"
           ]
         },
--- a/docs/json-schema/request.json
+++ b/docs/json-schema/request.json
@@ -93,6 +93,58 @@
       ],
       "type" : "object"
     },
+    "BuildCreateDevice" : {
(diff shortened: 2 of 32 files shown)


karenetheridge:
marking WIP pending the outcome of some design discussions, and more consideration given to deactivated plans.

PR #933, opened 2019-11-05
remove some redundant/unnecessary indexes

karenetheridge:
(see migrations 0107, 0135 in v3.0.0-a7)


karenetheridge:
deferred, to include in a subsequent PR.

PR #996, opened 2020-04-15
properly handle a report using a different device's system_uuid

karenetheridge:
If we do not query for the device properly, the wrong device will get updated;
this might also cause a database explosion if the system_uuid and/or
serial_number is already in use by another device.

also set both conflicting devices to the error state for better detection
    
This resolves Rollbar items 3490, 3491 occurring 2020-04-15.


karenetheridge:
Looks like buildbot is not working on this branch now.  I have run tests locally and all looks good.

PR #1005, opened 2020-06-16
new hardware_product endpoints for v3.0

### LINKED ISSUE #1003: hardware_product endpoints, for BOM visibility and manipulation
Needed for https://github.com/joyent/conch-ui/issues/213

- `GET /hardware_product` ought to only return the fields: id, sku, name, alias, generation_name, created, updated -- for the purpose of facilitating searches, or population of selection menus etc (fetching the entire table can be expensive as the 'specification' column especially can be huge, and this table might get big)

- new endpoint `PUT /hardware_product/:id_or_sku/specification` -- the 'path' query parameter in the URI represents a json pointer to the exact properties being set in the json blob (so the entire document does not need to be passed). for example:
`PUT /hardware_product/600-0042-001/specification?path=/disk_size/SEAGATE_8000` payload: `8192`  -> stores the integer `8192` in the specification field to so the document contains  `{"disk_size":{"SEAGATE_8000": 8192 } }` (other fields in the json are left untouched)
`PUT /hardware_product/600-0042-001/specification?path=/disk_size` payload: `{"SEAGATE_8000": 8192, "_default": 64 }` -- stores that json object at the "disk_size" property, overwriting all other properties under `/disk_size`

- `DELETE /hardware_product/:id_or_sku/specification` -- as the PUT endpoint above, takes a json pointer in the URI, to indicate which property should be deleted

karenetheridge:
new interfaces for direct manipulation of hardware_product.specification
    
    PUT /hardware_product/:hardware_product_id_or_other/specification?path=:json_pointer_to_data
    DELETE /hardware_product/:hardware_product_id_or_other/specification?path=:json_pointer_to_data

closes #1003.

### DIFF
--- a/docs/json-schema/common.json
+++ b/docs/json-schema/common.json
@@ -2,6 +2,47 @@
   "$comment" : "NOTE: This file is for human reference ONLY. For programmatic use, use the GET '/schema/common/$schema_name' endpoints, or within conch itself, json-schema/common.yaml.",
   "$schema" : "http://json-schema.org/draft-07/schema#",
   "definitions" : {
+    "HardwareProductSpecification" : {
+      "description" : "this is the structure of the hardware_product.specification database column",
+      "properties" : {
+        "chassis" : {
+          "properties" : {
+            "memory" : {
+              "properties" : {
+                "dimms" : {
+                  "$comment" : "items are in slot order, as in device report /dimms/*",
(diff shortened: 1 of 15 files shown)


karenetheridge:
cool. I'll get this up on conch-v3-staging today.

PR #1007, opened 2020-06-23
Allow changing the hardware_product for an occupied rack_layout...

### LINKED ISSUE #1002: Overwrite Rack Layout for Racks that have Assignments
We had a case just now where the rack layouts were populated with an outdated SKU. The devices that were put into the rack were correct so the SKU needed to be updated so that validations could proceed correctly.

To do this I had to:
1. capture the existing assignments
2. capture the existing layout
3. delete the existing assignments (with a JSON body specifying every device to be deleted)
4. update the existing layout data with the new hardware profile data
5.  restore the saved assignment data

The way that the preflight process works for the foreseeable future the assignments are less likely to change than the layout. It would be nice if we could simply update the layout without having it block on the existing assignment.

karenetheridge:
.. as long as the new hardware_product_id matches the occupying device (which
may require hitting POST /device/:device_id_or_serial_number/hardware_product
first). Other values (namely rack_unit_start) still cannot be changed for an
occupied layout.

closes #1002.

### DIFF
--- a/lib/Conch/Controller/RackLayout.pm
+++ b/lib/Conch/Controller/RackLayout.pm
@@ -173,10 +173,17 @@ sub update ($c) {
         return $c->status(400, { error => 'changing rack_id is not permitted' });
     }
 
-    # cannot alter an occupied layout
-    if (my $device_location = $layout->device_location) {
-        $c->log->debug('Cannot update layout: occupied by device id '.$device_location->device_id);
+    # only permit updating occupied layouts if the hardware_product_id is the only change,
+    # iff it is changing to the device_hardware_product_id.
+    if (my $device = $layout->related_resultset('device_location')
+        ->related_resultset('device')->columns([qw(id hardware_product_id)])->single) {
+      if (($input->{rack_unit_start} and $input->{rack_unit_start} != $layout->rack_unit_start)
+        or ($input->{hardware_product_id}
+          and $input->{hardware_product_id} ne $layout->hardware_product_id
+          and $input->{hardware_product_id} ne $device->hardware_product_id)) {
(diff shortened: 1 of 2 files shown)


lib/Conch/Controller/RackLayout.pm
```diff
@@ -173,10 +173,17 @@ sub update ($c) {
...
-    if (my $device_location = $layout->device_location) {
-        $c->log->debug('Cannot update layout: occupied by device id '.$device_location->device_id);
+    # only permit updating occupied layouts if the hardware_product_id is the only change,
+    # iff it is changing to the device_hardware_product_id.
+    if (my $device = $layout->related_resultset('device_location')
+        ->related_resultset('device')->columns([qw(id hardware_product_id)])->single) {
+      if (($input->{rack_unit_start} and $input->{rack_unit_start} != $layout->rack_unit_start)
+        or ($input->{hardware_product_id}
+          and $input->{hardware_product_id} ne $layout->hardware_product_id
+          and $input->{hardware_product_id} ne $device->hardware_product_id)) {
```
perigrin, line 183: Where is the device data populated from again? We pull it from the device reports no?

karenetheridge: in v3.0 the device is created during POST .../rack:Id_or_name/assignment, with a sku matching the rack_layout it is assigned to. The sku is not changed during device report processing.

If the sku needs to change after the device has been created, you need to do  POST /device/:device_id_or_serial/hardware_product (or .../sku).

So, to change the sku for a located device, change the device sku first, and then the rack layout sku can be changed to match it.

perigrin: Got it. That works for me.

PR #1029, opened 2020-09-30
drop unused column device_nic.iface_driver

karenetheridge:


perigrin (approved):

karenetheridge:
This has been fixed in kosh (https://github.com/joyent/kosh/pull/57) and I do not believe conch-ui is affected, so I'm merging this.

PR #1037, opened 2020-10-05
delete old device reports

karenetheridge:
We will delete all device reports (and associated validation_state records,
and then orphaned validation_results) that are older than 6 months, except we
retain the most recent report of each state (error, fail, pass).

This is implemented a database migration that runs when the commit is
deployed, as well as in a command that can be run at any time in the future.


karenetheridge:
Work is done for this but it cannot be deployed until the recovery effort for v2 validation data has completed running on production.

karenetheridge:
closed in favour of #1060.

PR #1053, opened 2020-10-26
move schema routes to /json_schema

karenetheridge:
- remove camel-casing of schema definition names (with a redirect for old names)
- use a more useful $id in returned schema definitions
- move all schema routes to /json_schema (with a redirect for the old endpoints)
- add GET /json_schema/common/:name and GET /json_schema/device_report/:name


perigrin (approved):
I think the changes on the shell are minimal, and only exist as of this week so no backwards compatibility issues.

PR #1054, opened 2020-10-28
simplify data returned by GET /build

karenetheridge:
- remove options ?with_device_health, ?with_device_phases, ?with_rack_phases
- remove 'completed_user' and 'admins' properties
- replace ?include_completed option with ?started=<0|1> and ?completed=<0|1>
for filtering results (by default, all builds are returned)

### DIFF
--- a/docs/json-schema/query_params.json
+++ b/docs/json-schema/query_params.json
@@ -68,19 +68,29 @@
       "$ref" : "#/definitions/ChangePassword"
     },
     "GetBuilds" : {
-      "allOf" : [
-        {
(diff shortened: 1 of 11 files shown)


bdelano (changes requested):
as far as I know we have 3 different 'status' for builds... so we should be able to see all or any. combination... please include all 3 and allow chaining
e.g.
`started=1&created=1` or `created=1&complete=1`

bdelano (dismissed):

perigrin (approved):
I'm a little sad to lose the "Completed By" on the full list, but it saves a join.

dustinryerson (approved):

